PR #6, opened 2026-03-12
Skip signout confirmation and redirect to home

santthosh:
## Summary
- Removes the NextAuth default signout confirmation page (the unstyled "Are you sure?" screen)
- Adds a `/signout` client page that immediately calls `signOut({ callbackUrl: "/" })` and redirects to the home page
- Updates all 7 server-side `redirect("/api/auth/signout")` calls across dashboard pages to use `/signout` instead

## Test plan
- [ ] Click sign out from the sidebar → should go straight to the home page with no confirmation
- [ ] Trigger a `TokenExpiredError` (e.g., revoke token) → should redirect to home, not the signout confirmation page
- [ ] `pnpm run build` passes (verified locally)


santthosh:
@mergewatch review can you run this again?

PR #19, opened 2026-03-19
Prevent duplicate review processing across containers

santthosh:
## Summary
- Adds atomic `claimReview()` to `IReviewStore` interface — prevents duplicate reviews when multiple containers or Lambda invocations process the same webhook
- **DynamoDB**: conditional `PutCommand` with `attribute_not_exists OR status IN (failed, skipped)`
- **Postgres**: `INSERT ... ON CONFLICT DO UPDATE WHERE status IN (failed, skipped) RETURNING`
- Both Lambda and server handlers call `claimReview()` before starting the pipeline — if another worker already claimed it, exits early

## Why this matters
- Without this, two containers behind a load balancer could both process the same PR webhook, running the full multi-agent pipeline twice (5-8 LLM calls each)
- Wastes LLM spend and can produce duplicate/conflicting GitHub comments
- Also protects against GitHub's webhook retry behavior

## Design decisions
- `claimReview()` allows re-processing of `failed` and `skipped` reviews (retriable states) but blocks `in_progress` and `complete`
- Uses database-level atomicity (DynamoDB conditional writes / Postgres conflict handling) — no external locks needed
- Existing `upsert()` is kept for backwards compatibility (used in skip path and other non-concurrent flows)

## Test plan
- [ ] `pnpm run build && pnpm run typecheck` passes
- [ ] Single webhook → review completes normally
- [ ] Duplicate webhook (same PR+commit) → second instance exits with "already in progress" log
- [ ] Failed review → retry succeeds (re-claims the record)
- [ ] Skipped review → re-review succeeds (re-claims the record)

### DIFF
--- a/packages/core/src/storage/types.ts
+++ b/packages/core/src/storage/types.ts
@@ -16,6 +16,13 @@ export interface IInstallationStore {
 
 export interface IReviewStore {
   upsert(review: ReviewItem): Promise<void>;
+  /**
+   * Atomically claim a review for processing.
+   * Inserts the review record only if no record with the same key exists
+   * or the existing record is not already in_progress/complete.
+   * Returns true if this caller claimed the review, false if another worker already has it.
+   */
+  claimReview(review: ReviewItem): Promise<boolean>;
   updateStatus(
     repoFullName: string,
     key: string,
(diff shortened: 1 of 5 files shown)


santthosh:
@mergewatch review, lets try again

PR #109, opened 2026-04-20
Add API keys tab to Settings

santthosh:
## Summary
- New dashboard page at Settings → API Keys for per-installation key management
- Route handler at `/api/api-keys` (GET list, POST create, DELETE revoke)
- Key generation: `mw_sk_live_` + base64url(24 random bytes); sha256 hash is persisted
- Raw key returned exactly once from POST; never exposed on GET
- Admin-check on mutation routes via existing GitHub helpers
- Uses the `DynamoApiKeyStore` from `@mergewatch/storage-dynamo`

## Test plan
- [x] Manual: create a key, confirm DynamoDB row with only the hash
- [x] Manual: raw key shown once, not on subsequent GETs
- [x] Manual: revoke removes the row
- [x] `pnpm run typecheck` passes


santthosh:
@mergewatch review

santthosh:
@mergewatch review

PR #145, opened 2026-05-19
feat(review): claim-aware grounding to kill false-positive criticals (W1+W2)

santthosh-mb:
## Problem

The diff-only review agents produce the highest-trust-damage class of false positive: a confident **CRITICAL** derived from a truncated hunk. The canonical case is the **"missing await"** finding flagged on a line that already reads `const x = await f()` — observed independently in **voice-bot #31** and **orca #39**, both times with a suggestion that *echoes the existing code*, and in #39 then marked "✅ Resolved" though nothing changed.

The existing structural grounding (`groundFinding`) couldn't catch these: it checks *identifier presence* within ±5 lines, and `migrationRunner(` / `searchViaPostgres(` **is** present near the anchor — so the false positive passed.

Implements **W1 + W2** from `docs/review-quality-plan.md` (PR #144).

## Changes

**W1 — deterministic no-op-suggestion guard** (`groundFinding`)
Drops a finding whose suggested code is already present in the file (whitespace-normalized). Splits the suggestion into code-shaped segments and requires **every** segment to already exist, so a not-yet-applied multi-line fix (e.g. a try/catch the code lacks) is preserved. Zero LLM cost — kills #31 deterministically.

**W2 — claim-aware critical verification** (`verifyCriticalFindings`)
Re-checks each surviving **critical** against the **complete** file with the light model, dropping it only on an explicit, parseable `valid:false`. **Fail-safe**: a missing file (couldn't fetch), an LLM error, or unparseable output **keeps** the finding — infrastructure trouble must never silently suppress a real critical. Criticals only (0–2 per PR), one light-model call each, run concurrently.

**Plumbing** — both stages share a single file fetch (`fetchFindingFileContents`) and run on a new `groundingFetch` context that is **not gated behind the `codebaseAwareness` flag** (verifying a critical against the real file is always worth the read; falls back to `fileFetchOptions`). Wired in both the Lambda and server handlers.

## Open questions from the plan — now answered by reading the code

- Agents get **diff only**; full-file fetch was opt-in behind `codebaseAwareness` → W2 decouples it.
- Grounding existed but was **identifier-presence only**, not claim-aware → that's exactly the gap W1/W2 close.
- `mergewatch triage` is **not parsed** anywhere (future W3).

## Verification

- New tests: **+13** (`suggestionAlreadyApplied`, no-op guard in `groundFinding`, `verifyCriticalFindings` incl. valid/invalid/missing-file/LLM-throw/unparseable/severity-scoping).
- `@mergewatch/core` **393 passed**, `@mergewatch/server` **75 passed**.
- `core` / `server` / `lambda` typecheck clean; full `pnpm run build` passes (12/12).

### DIFF
--- a/e2e/RUNBOOK.md
+++ b/e2e/RUNBOOK.md
@@ -147,6 +147,9 @@ Run these in order — they cover all current behaviors. ~30 minutes end-to-end.
 | [E2E-18](#e2e-18-delta-aware-verdict-on-security-improvement) | PR that resolves prior criticals → green verdict (≥4/5), not orange | 3m | 90s | tier-1 |
 | [E2E-19](#e2e-19-confidence-scores-hidden-by-default) | New install sees no `85%` etc. badges in finding rows | 30s | 60s | tier-1 |
 | [E2E-20](#e2e-20-pr-description-vs-code-drift-catch) | Stale "we now use X" in PR body → reviewer flags the mismatch | 2m | 60s | feedback |
+| [E2E-21](#e2e-21-no-op-suggestion-guard-w1) | Finding whose suggested fix already exists in the file → dropped | 1m | 60s | #145 |
+| [E2E-22](#e2e-22-claim-aware-critical-verification-w2) | "Missing await" critical on code that already awaits (truncated-diff artifact) → dropped by full-file verification | 1m | 60s | #145 |
+| [E2E-23](#e2e-23-re-review-convergence--no-whack-a-mole-w9w3) | Re-review never reports the same finding as both "✅ resolved" and "🆕 new"; a triage-rebutted finding is not re-raised | 3m | 90s | #145 (target: W9+W3) |
 
 ---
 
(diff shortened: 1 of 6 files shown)


santthosh-mb:
## mergewatch triage (fixed in latest commit)

**⚠️ Concurrent LLM calls / rate-limiting (`verifyCriticalFindings`)** — **fixed.** Genuine catch. It used unbounded `Promise.all` over all criticals; a pathological PR with many criticals would burst N parallel Bedrock `InvokeModel` calls and hit the per-minute TPM quota — the exact reason `runReviewPipeline` already routes agents through `withConcurrency(..., AGENT_CONCURRENCY)`. Switched the verification fan-out to the same helper. `withConcurrency` preserves input order, so the `verdicts[i] ↔ findings[i]` alignment is unchanged.

**⚠️ Silent failure on unparseable JSON (default fallback)** — **fixed.** The fail-safe (keep the critical when the verdict can't be parsed) is correct and stays, but it was silent. Now uses a `{}` sentinel default instead of `{ valid: true }` and logs `no usable verdict — keeping finding (fail-safe)`, so the kept-on-bad-output path is observable. This also covers the legitimate part of the "broad catch" finding.

**⚠️ "Broad exception catching masks failure modes"** — **rebutted.** The catch-all is the intentional fail-safe: infra trouble must never silently drop a real critical. Every path inside the `try` throws the same `Error` out of `llm.invoke` (network/timeout/throttle are not separable without provider-specific typing), so per-type branching would be guesswork. The real observability gap — *why* a finding was kept — is now logged (above).

**⚠️ "Potential code injection / ReDoS in `suggestionAlreadyApplied`"** — **framing is a false positive; cap added anyway for consistency.** The regexes here (`` /```[a-zA-Z]*\n?|```/g ``, `/\n|(?<=:)\s/`, `/^[`'"\s]+|[`'"\s.]+$/g`, `/\s+/g`) are all linear — no nested quantifiers, no catastrophic backtracking — and the suggestion is LLM-generated, not request-controlled, so there is no injection/ReDoS path. But the codebase already has a `FINDING_TEXT_MAX_BYTES = 4096` ReDoS-defense convention (`extractFindingIdentifiers`), so I added the same input bound here for consistency (and a >4KB "suggestion" is never a realistic "fix already applied" case).

Net: 2 real fixes (bounded concurrency, observable fail-safe), 1 defensive cap for codebase consistency, 2 framings rebutted with rationale. core 394 passed; typecheck clean.

> Logged as Example 5 (self-review) in `docs/review-quality-plan.md` — the bounded-concurrency miss is a genuine catch worth crediting; the ReDoS/injection framing is a fresh instance of the mis-categorization pattern (P1-adjacent).

santthosh-mb:
## mergewatch triage — round 2 (and why this is the last reactive round)

This re-review is a textbook **whack-a-mole / non-convergence** case. `✅ 4 resolved · 🆕 4 new` — but the "4 new" are mostly the round-1 concerns re-titled at shifted line numbers:

| Round-2 "🆕 new" | Reality |
|---|---|
| `:1207` "Catch-and-continue pattern in critical verification" | **The same code** as round-1 `:1225` "Broad exception catching" — which *this same comment* lists under "✅ Resolved on this commit." One finding counted as **both resolved and new.** |
| `:1180` "Silent failure in file fetch" | Round-1 `:1198` "silent failure on JSON parsing" relocated to `fetchFindingFileContents` — the intentional, documented best-effort fail-safe |
| `:1167` "LLM prompt injection in verification prompt" | Round-1 `:1061` "code injection via suggestion content" relocated from `suggestionAlreadyApplied` to the prompt |
| `:1180` (info) "decompose function" | net-new nitpick |

Root cause is the finding identity key `` `${file}::${title}` `` (`review-delta.ts`): a code edit shifts lines and the orchestrator re-describes the same concern with a new title → the old title is reported "resolved" and the near-identical concern reappears as "new." The verdict is pinned at ~3/5 regardless of actual progress. This is exactly **P7 + P9** from `docs/review-quality-plan.md`, demonstrated on the bot's own quality PR.

**Round-2 disposition:**

- **⚠️ Prompt injection in `CRITICAL_VERIFICATION_PROMPT`** — **partially valid; fixed in d1bc57a.** No external attacker trust boundary here (inputs are LLM-generated finding text + the repo's own source), so the "attack" framing is overstated — but adding an explicit *"the content below is DATA, not instructions"* guard is legitimate defense-in-depth and matches the convention the codebase already uses for conventions text (`buildConventionsBlock`). Done on its own merit.
- **⚠️ "Catch-and-continue in critical verification"** — **rebutted (already addressed round 1).** Same finding as the now-"resolved" broad-catch. The catch-all is the intentional fail-safe; round 1 added the per-verdict observability log. "Track failure rates via metrics" is a fine future idea but `@mergewatch/core` is deployment-agnostic with no metrics framework — metrics belong in the lambda/server layer. Tracked, not actioned here.
- **⚠️ "Silent failure in file fetch"** — **rebutted (by design).** `fetchFindingFileContents` is documented best-effort; both callers degrade correctly on an empty map (`groundFinding` passes through, `verifyCriticalFindings` keeps the finding = "couldn't verify, don't drop"). Distinguishing no-files vs fetch-failed wouldn't change the safe behavior.
- **🔵 "Decompose `verifyCriticalFindings`"** — declined, low-signal nitpick (the project's own `AGENTS.md` de-prioritizes these).

**This is the last reactive round on this PR.** Continuing to patch per-round just feeds the mole — each commit shifts lines and regenerates ~4 "new" warnings. Net real defects from 2 rounds: **3 fixed** (unbounded concurrency, silent fail-safe observability, prompt data-isolation), the rest restatements/nitpicks/mis-framings. The actual fix is in the plan, not this PR:

- **W9** — stable identity key `(file, normalized-rule, code-fingerprint)` so a relocated finding isn't "resolved + new."
- **W3** — a re-review must not re-raise a `(file, rule)` the author already rebutted in a `mergewatch triage` reply on a prior commit (convergence guard).

Logged as round-2 evidence in Example 5 of the plan. #145 is advisory 3/5 with **no criticals** — not merge-blocked.

PR #163, opened 2026-05-21
FP-G — linter-aware style agent (final item on the FP-reduction plan)

santthosh-mb:
## Summary

The **last item** on the false-positive-reduction plan (after #159 / #160 / #161 / #162). `STYLE_REVIEWER_PROMPT` has always carried the rule *"Anything already enforced by a linter — DO NOT report"*, but the model had no way to know which linters the repo had configured. It conservatively re-reported semicolons / `prefer-const` / unused-import / quote-style nits because it didn't trust ESLint / Biome / Ruff / etc. would catch them. FP-G closes that gap deterministically.

## Detection
`packages/core/src/config/conventions.ts`

- New `detectLinters(octokit, owner, repo, ref)` + `DetectedLinter` type.
- Single root-listing API call (`repos.getContent` with `path: ''`), matches root entries against per-linter marker tables in `LINTER_MARKERS`:

| Linter | Markers |
|---|---|
| `eslint` | `.eslintrc`, `.eslintrc.{js,cjs,mjs,json,yml,yaml}`, `eslint.config.{js,ts,mjs,cjs}` |
| `biome` | `biome.json`, `biome.jsonc` |
| `ruff` | `ruff.toml`, `.ruff.toml`, plus `pyproject.toml` when it contains `[tool.ruff…]` |
| `flake8` | `.flake8` |
| `clippy` | `clippy.toml`, `.clippy.toml` |
| `golangci` | `.golangci.yml`, `.golangci.yaml`, `.golangci.toml` |
| `stylelint` | `.stylelintrc`, `.stylelintrc.{json,js,cjs,yml,yaml}`, `stylelint.config.{js,cjs}` |

- Returns the set sorted lexicographically (deterministic for prompt caching).
- Best-effort: any API error returns \`[]\`. **Root listing only** — \`node_modules/.eslintrc.json\` doesn't trigger the directive.

## Prompt + pipeline + handlers

- `STYLE_REVIEWER_PROMPT` gains `LINTER_AWARE_PLACEHOLDER` (`{{LINTERS_DETECTED}}`); `buildLinterAwareDirective(linters)` renders the deferral directive or returns \`\`\`\`\`\` (placeholder stripped → prompt byte-identical to pre-FP-G).
- `runStyleAgent` accepts optional `detectedLinters?: readonly string[]` and substitutes the placeholder. **Style-agent only** — other agents are byte-identical.
- `ReviewPipelineOptions.detectedLinters` plumbs from both handlers.
[...]

PR #166, opened 2026-05-22
FB-C + FB-D — inline-comment reactions + /mergewatch reject

santthosh-mb:
## Summary

**PR 2 of the FP-feedback plan** (#164). Captures explicit per-finding FP signal from GitHub-native channels:
- **Passive** — 👎 / 🤔 reactions on bot inline comments (no typing needed)
- **Active** — `/mergewatch reject <category>` slash command (categorical signal)

Both feed the `FindingDispositionRecord` counters laid down by FB-A. No new UI surface yet — that's PR 4+.

## FB-C — Inline-comment reactions → disputes / agreements

`packages/core/src/insights/disposition-writer.ts` + handlers

- New `pollAndRecordInlineReactions(octokit, …)`. Single `pulls.listReviewComments` call — the summary `reactions` field on each comment already carries per-type counts, so no per-comment reaction-list API call is needed.
- **Reaction mapping** (locked):

  | Reaction | Counter |
  |---|---|
  | 👎 `-1` | `disputeCount` |
  | 🤔 `confused` | `disputeCount` |
  | 👍 `+1` | `agreementCount` |
  | ❤️ `heart` | `agreementCount` |
  | 🚀 `rocket` | `agreementCount` |
  | 👀 / 🎉 / 😄 | ignored (read-receipt / ambiguous) |

- **Monotonic via snapshot**: only the positive delta vs the prior snapshot triggers increments. A reviewer who 👎 then *removes* the reaction can't make the dispute counter regress.
- **Filter to MergeWatch comments only**: bot-authored AND carrying `INLINE_BOT_COMMENT_MARKER`. CopilotAI / dependabot / human inline comments stay out.
- **Storage**: new `ReviewItem.inlineReactionsSnapshot` field (jsonb on Postgres via migration `0004_…`; DynamoDB is schemaless).
- **Wiring**: both handlers fold the poll into the existing post-pipeline analytics block as FB-A/B. New snapshot persists via `updateStatus`.
- **Best-effort**: a failed `listReviewComments` returns an empty snapshot and logs `[fb-c] failed to list review comments` — never blocks a review.

## FB-D — `/mergewatch reject` slash command

`packages/core/src/agents/inline-reply.ts` + handlers

- New `parseRejectIntent(text)` recogniser. Grammar (locked in PR #164):

  ```
  /mergewatch reject <category> [optional free-text]
  ```

- **Categories** (closed set): `already-handled`, `out-of-scope`, `wrong-target`, `style-disagreement`, `other`.
- **Silent-other coercion**: anything unrecognised coerces to `other` and preserves the typed token as the leading word of the free-text reason. Bare `/mergewatch reject` (no category) also coerces to `other` with no text. *Preserves the signal even when the reviewer mistypes.*
- `handleInlineReply` gained a new `'rejected'` action between resolve and the LLM-reply branch. Posts a confirming bot reply naming the recorded category (and explaining the coercion when it happened).
- **Does NOT auto-resolve** the GitHub thread — rejection records the signal; closure is a human decision. `/resolve` and `/reject` stay orthogonal verbs. (Precedence: `/resolve` wins when both intents appear in the same reply.)
- **`InlineReplyResult`** gained `rejectedFindingKeys`, `rejectCategory`, `rejectText`. Handlers translate these into:
  - `appendRejectReason` on each match key (categorical analytics)
  - `recordDisputes` on each match key (dispute count)

## Tests

| File | New cases | Coverage |
|------|-----------|---------|
| `core/src/agents/inline-reply.test.ts` | **13** | `parseRejectIntent` — every category, case-insensitive, mistyped coercion, bare verb, prose-form rejection, multi-line scope. `handleInlineReply` rejected-action — path with category + text, silent-other reply explainer, resolve-takes-precedence, no-keys fallback. |
| `core/src/insights/disposition-writer.test.ts` | **11** | `pollAndRecordInlineReactions` — each mapping branch, monotonic-via-delta, reaction-removal no-op, non-bot skip, missing marker skip, missing path → snapshot-only, empty store no-op, snapshot return shape, API failure → empty snapshot. |

Local validation:
- `core test` **556 / 556** (was 532 + 24 new)
- `core typecheck` clean
- `drizzle migrations:check` 🐶🔥
- `pnpm run build` 12/12
- `pnpm run typecheck` 20/20
- `pnpm run test:coverage` **20/20**

## Migration safety

- **Postgres 0004**: `ADD COLUMN IF NOT EXISTS` (per CLAUDE.md).
- **DynamoDB**: schemaless; new `inlineReactionsSnapshot` surfaces without DDL.
- **Old reviews** without the snapshot get treated as "everything is new" on first poll — first poll for an established PR may emit a burst of dispute/agreement increments (catch-up). Acceptable for v1; the counters are monotonic so over-counting is the only risk, not undercounting.

## E2E Regression

`e2e/RUNBOOK.md` — **E2E-39** + **E2E-40** flipped from TARGET to ✅ SHIPPED, per [[e2e-regression-spec-per-pr]].

## Plan doc

`docs/false-positive-feedback-plan.md` — FB-C + FB-D marked **✅ SHIPPED** in section headings and the priority table.

## Next

**PR 3** in the plan is **FB-E** (nightly `InstallationFPInsight` rollup) — unblocked once #166 merges. After that the dashboard charts (PR 4+) can finally read aggregated data.

### DIFF
--- a/docs/false-positive-feedback-plan.md
+++ b/docs/false-positive-feedback-plan.md
@@ -157,7 +157,7 @@ Ranked by dependency order (foundation first → capture → aggregation → sur
 
 ---
 
-### FB-C — Inline-comment 👎 reactions → disputes
+### FB-C — Inline-comment 👎 reactions → disputes  ✅ SHIPPED
 
 **Where the gap lives:** Top-comment reactions are already collected via `ReviewItem.reactions`. Inline-comment reactions are not. Yet reactions on an inline finding are arguably *more* signal-dense than top-level reactions — they're per-finding, attributed to the reviewer, and require zero typing.
 
@@ -172,7 +172,7 @@ The bot's inline-comment ID is already linked to a finding via `ReviewItem.findi
 
 ---
 
-### FB-D — `/mergewatch reject "<reason>"` slash command
+### FB-D — `/mergewatch reject "<reason>"` slash command  ✅ SHIPPED
 
 **Where the gap lives:** W3 triage prose is the cleanest signal we have — author writes a sentence explaining why a finding is wrong — but it requires the author to write prose and lives only on top-level comments. Inline threads have only `/resolve` (which is success-shaped, not FP-shaped). There's no quick "this finding is wrong" channel from an inline thread.
 
@@ -302,8 +302,8 @@ Compute cost is bounded by the largest installation's record count; rollups stay
 |---|---|---|---|---|---|
 | **FB-A** | FindingDispositionRecord storage + writers | Persist | S | — | ✅ SHIPPED |
 | **FB-B** | Quiet-drop derived counter | Persist | S | FB-A | ✅ SHIPPED |
-| **FB-C** | Inline-comment 👎 → disputes | Capture | M | FB-A | ★★★ |
-| **FB-D** | `/mergewatch reject` slash command | Capture | M | FB-A | ★★ |
+| **FB-C** | Inline-comment 👎 → disputes | Capture | M | FB-A | ✅ SHIPPED |
+| **FB-D** | `/mergewatch reject` slash command | Capture | M | FB-A | ✅ SHIPPED |
 | **FB-E** | Nightly InstallationFPInsight rollup | Aggregate | M | FB-A, FB-B | ★★★ |
 | **FB-F** | FP funnel chart | Surface | M | FB-E | ★★ |
 | **FB-G** | Dispute-rate-by-agent chart | Surface | M | FB-E | ★★ |
--- a/e2e/RUNBOOK.md
+++ b/e2e/RUNBOOK.md
@@ -165,8 +165,8 @@ Run these in order — they cover all current behaviors. ~30 minutes end-to-end.
 | [E2E-36](#e2e-36-fp-g--linter-aware-style-agent) | Repos with detected linters (eslint / ruff / clippy / biome) get a stricter STYLE_REVIEWER_PROMPT that defers lint-equivalent findings (FP-G) | 2m | 60s | FP-G |
 | [E2E-37](#e2e-37-fb-a--findingdispositionrecord-storage--writers) | FindingDispositionRecord rows are written on every surfacing, W3 dispute, FP-F inline-resolve (FB-A) | 2m | 60s | FB-A |
 | [E2E-38](#e2e-38-fb-b--quiet-drop-derived-counter) | Quiet-drop (finding gone without code change) increments `silentDropCount` on the matching record (FB-B) | 2m | 60s | FB-B |
-| [E2E-39](#e2e-39-fb-c--inline-comment--reactions--disputes-target) | 👎 / 🤔 on a bot inline comment increments `disputeCount`; 👍 / ❤️ / 🚀 increments `agreementCount` (FB-C) — **TARGET** | 2m | 60s | FB-C |
-| [E2E-40](#e2e-40-fb-d--mergewatch-reject-slash-command-target) | `/mergewatch reject <category> [reason]` on an inline thread persists a categorised rejection + posts a confirming bot reply (FB-D) — **TARGET** | 3m | 90s | FB-D |
+| [E2E-39](#e2e-39-fb-c--inline-comment--reactions--disputes) | 👎 / 🤔 on a bot inline comment increments `disputeCount`; 👍 / ❤️ / 🚀 increments `agreementCount` (FB-C) | 2m | 60s | FB-C |
+| [E2E-40](#e2e-40-fb-d--mergewatch-reject-slash-command) | `/mergewatch reject <category> [reason]` on an inline thread persists a categorised rejection + posts a confirming bot reply (FB-D) | 3m | 90s | FB-D |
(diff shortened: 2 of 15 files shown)


santthosh-mb:
Triaged the MergeWatch review in commit ca3e66c.

**Applied** ✅

1. **Hardcoded `RejectCategory` union duplicated** (info #4 + #5) — replaced both inline unions in `server/review-processor.ts:197` and `lambda/handlers/review-agent.ts:263` with `NonNullable<FindingDispositionRecord['rejectReasons']>[number]`. Now the inline shape can't drift if the union is widened — both handlers reference the canonical exported type from `@mergewatch/core`.

2. **Silent `appendRejectReason` failures** (warnings #2 + #3) — replaced the per-key sequential `for ... await ... .catch()` loop in both handlers with `Promise.allSettled` + a single summary log line:
   ```
   [fb-d] N/M appendRejectReason write(s) failed for owner/repo#42 (category=...)
   ```
   Three wins: parallel I/O, one greppable log line per batch instead of N, failure rate visible at a glance. **Kept the fail-open posture as designed** — analytics writes never block the review path. Not re-throwing or adding a retry mechanism: a retry would slow the review for a temporary outage while still risking eventual consistency. The summary log is sufficient to drive a future CloudWatch alarm.

**Intentional design** 🟡

3. **`listReviewComments` API failure visibility** (warning #1) — the existing `[fb-c] failed to list review comments for owner/repo#42` log line is already structured (prefixed for grep) and flows through `console.warn` → CloudWatch, which can alarm on the prefix directly. Adding in-code metric emission would duplicate that signal. If we observe recurring failures in the wild, the right move is a CloudWatch alarm on `[fb-c] failed to list` — not extra code.

4. **GitHub API response validation** (info #6) — Octokit already validates against the GitHub OpenAPI spec at the SDK boundary, and our code does explicit type coercion (`Number(…)`, `String(…)`, `typeof c.path === 'string'`) on every field we read. Adding hand-rolled schema validation on top would be redundant and risk drift if GitHub adds fields.

Core 556/556 still green; `pnpm run build` 12/12; `pnpm run test:coverage` 20/20.

santthosh-mb:
Reviewed the round-2 findings carefully — **all three are false-positive misreads**, but one of them did surface a real consistency improvement I applied in 842543f.

### Detailed triage

**Warning #1 main (`:262` — "doesn't log error details")** — false positive
Line 262 is literally:
```ts
console.warn('[fb-c] failed to list review comments for %s/%s#%d:', owner, repo, prNumber, err);
```
The `err` IS the trailing argument. The reviewer's suggested fix is identical to my code. The model didn't recognise that `console.warn` accepts trailing args after a format string.

**Warning #1 cluster siblings (`:281`, `:290`)** — misanchored
- Line 281 is `if (n > 0) currentForComment[type] = n;` — assignment, no error handling.
- Line 290 is `const matchKeys = findingMatchKeys(…);` — also no error handling.

The W10 cluster aggregated three unrelated lines into one finding.

**Warning #2 (`:209` JSDoc says only 'replied')** — false positive, wrong field
The reviewer is reading the JSDoc on line 191 — which is attached to `recommendation` (line 192). And `recommendation` IS correctly scoped to `'replied'` only (it's the LLM-reply recommendation; not set on `/reject`).

The actual `botCommentId` JSDoc is on line 193 (right above line 194 where `botCommentId` is declared) and correctly states `Populated when action === 'replied' or 'rejected'`. Both branches DO set it. No drift.

**Info #3 (data validation)** — same as round-1
Octokit validates against GitHub's OpenAPI spec at the SDK boundary; the code does explicit `Number()` / `String()` / `typeof` coercion on every field. Hand-rolled schema validation would duplicate the SDK's work.

### What I did apply (842543f)

The W10 cluster, despite being misanchored, contains a real *substantive* point: the per-reaction-per-key increment loop in `pollAndRecordInlineReactions` still used the same per-call sequential `await + try/catch + continue` pattern I refactored for `appendRejectReason` in round-1. Refactoring it to `Promise.allSettled` + single summary log for **consistency** with the appendRejectReason batch — not because the line pointer was correct, but because the pattern itself is worth applying uniformly across all "best-effort fan-out of analytics writes".

```ts
const incrementOps: Promise<void>[] = [];
for (const c of comments) { /* … */
  for (let i = 0; i < delta; i++) for (const key of matchKeys) {
    incrementOps.push(increment(inst, repoFullName, key));
  }
  /* … */
}
if (incrementOps.length > 0) {
  const settled = await Promise.allSettled(incrementOps);
  const failed = settled.filter((r) => r.status === 'rejected').length;
  if (failed > 0) console.warn('[fb-c] %d/%d reaction increment write(s) failed on %s', failed, settled.length, repoFullName);
}
```

Same shape as the FB-D batch — one pattern for the whole analytics fan-out layer now.

### Meta-observation

This is the **first** MW review on this branch that produced findings the underlying code didn't actually exhibit — interesting data point for the FP-feedback work we're building. Post-FB-A merge (#165), these would all increment `disputeCount` once we add the dispositions table to production. Once FB-C ships, MW's own follow-up review at #163 (4 findings, all valid) vs this one (3 findings, 0 valid) would surface as data points in the dispute-rate-by-agent chart (FB-G) — the "review quality" agent's recent dispute rate.

Core 556/556 still green; `pnpm run build` 12/12; `pnpm run test:coverage` 20/20.

PR #169, opened 2026-05-22
FB-E — nightly InstallationFPInsight rollup (FP-feedback PR 3)

santthosh-mb:
## Summary

**PR 3 of the FP-feedback plan** (#164). Aggregates the per-finding `FindingDispositionRecord` rows (FB-A) into per-installation **rolling-window rollups** (7d / 30d / 90d) that the dashboard charts (PR 4+: FB-F..FB-J) will read. Closes the **persist → aggregate** arc.

No UI surface yet — pure batch job. Once this merges, the dashboard work is unblocked.

## Storage layer

- New `InstallationFPInsight` type (`packages/core/src/types/db.ts`):
  - `totalFindingsSurfaced / totalDisputes / disputeRate`
  - `totalSilentDrops / totalAgreements`
  - `perCategory: Record<string, { surfaced, disputed, rate }>` — for FB-G
  - `perRepo: Record<string, { surfaced, disputed, rate }>` — for FB-J
  - `topClusters: Array<{ sigTokens, representativeTitle, surfaceCount, disputeCount, rate }>` — for FB-H
- New `IFPInsightStore` interface — `upsert / get / listByInstallation`.
- New `IInstallationStore.listInstallationIds()` — rollup needs to fan out across every installation.
- **Postgres**: new `installation_fp_insights` table + migration `0005_acoustic_puma.sql` (`CREATE TABLE IF NOT EXISTS`).
- **DynamoDB**: new `mergewatch-installation-fp-insights-${Stage}` table in `infra/template.yaml`. Composite PK `(installationId, window)` so one `Query` returns all three windows. `DeletionPolicy: Retain`.
- `disputeRate` stored as text in both backends — same precision discipline as `reviews.estimated_cost_usd`, avoids float drift across re-reads.

## Aggregation algorithm

New pure function `buildInsightFromDispositions(installationId, window, windowEndIso, records)`:

1. Filter records by `lastSeen ∈ [windowEnd - windowLength, windowEnd]`.
2. Sum global counters; bucket by `category` + `repoFullName`.
3. **Cluster** via union-find on shared `sigTokens` (≥1 token overlap unions records). Reuses the W10 token bag captured on every surfacing by FB-A.
4. Representative title = highest-surfacing member's title parsed from `findingMatchKey`. Falls back to file path for fingerprint-form keys.
5. Sort `topClusters` by `disputeRate × surfaceCount` (leverage); cap at 10.

Pure function — no storage / time / I/O coupling. Fully unit-testable.

[...]

PR #172, opened 2026-05-22
FB-F + FB-G — FP funnel + dispute-rate-by-agent on /dashboard/insights

santthosh-mb:
## Summary

**PR 4 of the FP-feedback plan** (#164). First **user-visible surface** of the data we've been collecting since PR 1. New `/dashboard/insights` route with two charts wired to the FB-E rollup. **FB-I (severity-shopping) re-scoped** to a follow-up PR — see *Re-scoping* below.

## What you get

**`/dashboard/insights`** — new sidenav entry between *Analytics* and *Repositories*:

- **Window selector** — 7d / 30d / 90d toggle. Drives both charts.
- **FB-F: FP funnel** — horizontal stacked bar over `totalFindingsSurfaced`. Four segments: `unsignaled` (no signal) + `agreed` (👍/❤️/🚀) + `silentDropped` (implicit FP) + `disputed` (explicit FP). Sums to total by construction. Tooltip shows percentages.
- **FB-G: Dispute rate by agent** — horizontal bar per `perCategory` entry. Severity colouring: 🔴 ≥ 50%, 🟠 ≥ 25%, 🟦 otherwise. Tells the operator which agent is the noisiest for *this* org.
- **Zero-state panel** when the rollup hasn't produced rows yet (fresh installation) OR the dashboard's `fpInsights` store isn't wired (older deployments without `DYNAMODB_TABLE_FP_INSIGHTS` set).

## v1 shape decisions

| Original spec | What v1 ships | Why |
|---|---|---|
| Funnel: `surfaced → carried → resolved → disputed → silently-dropped` | Four-segment stacked bar over surfaced count | "Carried" + "resolved" need a finding-state machine the rollup doesn't yet have; v1 ships the four signals we actually track in `FindingDispositionRecord` |
| FB-G: line chart over time, one line per agent | Horizontal bar of `perCategory` rates in active window | True time-series needs per-day rollup buckets the FB-E job doesn't emit (one snapshot/night with 7d/30d/90d sliding windows). Operator can compare windows manually via the selector. Upgrade when FB-E gains per-day mode. |

Both are documented in the runbook + plan doc with the trigger conditions for the time-series upgrade.

## Data plumbing

- New `IDashboardFPInsightStore` interface in `@mergewatch/core`. Optional field on `IDashboardStore` so older deployments keep working — chart routes render zero-state when absent.
- **DynamoDB**: `createDynamoDashboardStore` accepts optional `fpInsightsTable`; reuses the FB-E store instance.
- **Postgres**: factory wires the store unconditionally (Drizzle schema in the same package, no per-deployment config).
- `dashboard/lib/store.ts` threads the new env var `DYNAMODB_TABLE_FP_INSIGHTS`.
- `next.config.js` `env` block exposes it (Amplify SSR quirk per CLAUDE.md).
- SAM template `Outputs` surfaces the new table name + the FB-A FindingDispositions table name (was missing).

## API route

`GET /api/insights?installation_id=…` — same access-gate pattern as `/api/settings`. Returns `{ insights: InstallationFPInsight[] }` or `{ insights: [] }` on any soft-fail. Never 500s on the happy path.

## Re-scoping: FB-I → follow-up PR

`FindingDispositionRecord` only tracks `category` (agent name), not `severity`. Building the FB-I severity-shopping detector needs:
1. Add `severity` field to `FindingDispositionRecord`.
2. Migrate Postgres + DynamoDB.
3. Update `recordFindingSurfacings` writer to set it from `finding.severity`.
4. Add `perSeverity` bucket to `InstallationFPInsight` rollup.
5. Build the FB-I chart.

That's a self-contained ~M-effort follow-up PR. Re-scoped + documented in:
- `docs/false-positive-feedback-plan.md` → FB-I marked ⏸ DEFERRED with trigger condition.
- `e2e/RUNBOOK.md` → E2E-45 stays TARGET.

## Tests + validation

- **Local**: `core test` 594/594; `pnpm run build` 12/12; `pnpm run typecheck` 20/20; `pnpm run test:coverage` 20/20.
- **Dashboard build**: clean (Next 15 + recharts already present in the dep tree).

UI is React + recharts; no dashboard-test-harness in the repo (per `AGENTS.md`), so no test files added on the dashboard side. Validation came from `pnpm --filter @mergewatch/dashboard run build` succeeding end-to-end.

## Migration safety

- **Postgres**: `installation_fp_insights` migration shipped in #169. No new migration here.
- **DynamoDB**: `DYNAMODB_TABLE_FP_INSIGHTS` is optional in v1. Setting it via SAM (this PR adds the Output) makes it visible; absence is a no-op (zero-state).
- **No breaking change** to existing routes — `IDashboardStore.fpInsights` is `?`.
- Sidenav nav entry is new but additive — existing routes unchanged.

## E2E Regression

- **E2E-42** flipped from TARGET → ✅ SHIPPED with v1-shape note.
- **E2E-43** flipped from TARGET → ✅ SHIPPED with v1-shape note.
- **E2E-45** (FB-I) stays TARGET — re-scoped to follow-up.

## Plan doc

`docs/false-positive-feedback-plan.md` — FB-F + FB-G marked ✅ SHIPPED in section headings; FB-I marked ⏸ DEFERRED with trigger conditions; priority table updated.

## Next

**PR 5** in the plan is **FB-H + FB-J + FB-K** (top recurring themes table + per-repo heatmap + `.mergewatch.yml` rule-suggestion CTA). All three read from the same rollup data the charts above use — no new schema work needed.


santthosh-mb:
Triaged in commit 00dd30d.

### Applied ✅

1. **`:14` stale route comment** — refactored from "insights as analytics subview" to "insights as own route" mid-implementation; the API route's JSDoc didn't follow. Now matches the page.tsx header doc.
2. **`:96` defensive clamping on funnel** — the math is provably non-negative by induction from the earlier Math.min clamps, but the reviewer's suggestion to wrap each step in Math.max(0, ...) is cheap protection against a future rollup-aggregation refactor that quietly breaks the invariant. Applied as belt-and-braces.

### Disagree 🟡

3. **`:42` "SQL injection via unvalidated installation_id"** — false positive. Both backends parameterize:
   - Postgres: \`eq(installationFpInsights.installationId, installationId)\` via Drizzle ORM
   - DynamoDB: \`KeyConditionExpression: 'installationId = :id'\` via AWS SDK placeholder
   
   No string interpolation reaches the query. The reviewer pattern-matched on "user input → store call" without modelling the ORM / SDK abstraction layers.
4. **`:122` "URL injection via unvalidated installationId prop"** — false positive. \`installationId\` is a server-side-resolved prop from \`activeInstallation.id\` (a number from GitHub's API), not user input. Plus \`encodeURIComponent\` is already applied AND \`/api/insights\` re-validates membership server-side. Adding a regex check would be cargo-cult defense-in-depth on a non-existent threat surface.
5. **`:58` "silent failure masks access control"** — design choice. Returns \`{ insights: [] }\` on flaky GitHub-API failure rather than 500ing the dashboard page. Matches the existing \`/api/settings\` route pattern. The payload contains no actual user data; the failure direction is safe.

### Meta take

This is the **first PR reviewed after FP-H/I/J landed in main** (#171). Worth checking whether those guards activated — they didn't, because **this is a first review, not a re-review**. FP-H/I/J L2 only fire when \`previousFindings\` is non-empty.

What DID happen: 3 of 5 false positives are **generic-pattern hallucinations** — the model sees a shape ("user input → store call", "user input → URL", "subtraction result") and emits the canonical "X-injection / negative-value" finding without verifying the actual code path:
- "SQL injection!" → didn't check the ORM
- "URL injection!" → didn't trace the prop's source
- "Negative value!" → didn't verify the clamp chain

This is a class FP-H/I/J **don't address**. It's not re-review-specific; it's a first-review failure mode where the LLM short-circuits abstraction-aware verification.

**Possible new workstream — FP-K: abstraction-aware verifier**

Extend \`FINDING_VERIFICATION_PROMPT\` (already touched by FP-E, FP-I, FP-J) with an explicit "known-safe abstractions" list:

> *"Common safe abstractions (mark INVALID if the suggested injection / unvalidated-input concern goes through any of these):*
> *- Drizzle ORM \`eq(col, value)\` and similar query builders parameterize all values.*
> *- AWS SDK \`ExpressionAttribute*\` placeholders parameterize all values.*
> *- React's automatic JSX escaping prevents XSS on rendered strings.*
> *- \`encodeURIComponent\` on a fetch URL prevents URL injection.*
> *- Prepared statements / parameterized queries via any ORM."*

This would catch all three pattern-matched false positives from this round in one prompt extension. Same shape as FP-H L2 + FP-I L1 + FP-J L2 — all live inside the verifier prompt — so the cost is just more INVALID conditions in the same single LLM call per finding.

**Want me to file it as #173 + bundle it with a future verifier-prompt PR?** Or wait until we see this pattern again? My read: file it now while the evidence is fresh — three out of five findings on this PR alone justify the workstream existing.

### Hit-rate update

| PR | Score | Visible findings | Valid | Hit rate |
|---|---|---|---|---|
| #163 (FP-G) | 3/5 | 4 | 4 | 100% |
| #166 r1 (FB-C+D) | 3/5 | 6 | 4 | 67% |
| #166 r2 | 3/5 | 3 | 0 | 0% |
| #169 r1 (FB-E) | 2/5 | 3 | 1 | 33% |
| #169 r2 | 3/5 | 2 | 1 | 50% |
| **#172 r1 (FB-F+G)** | 3/5 | 5 | 1 (+1 defensive) | **20–30%** |

Trend is still noisy. Round-1 hit rates: 100% → 67% → 33% → 20%. Worth tracking via FB-G's chart once the data accumulates — this is exactly what FB-G is for, just on first-reviews now rather than only re-reviews.

\`pnpm run build\` 12/12; \`pnpm run test:coverage\` 20/20.

santthosh-mb:
Triaged in commit \`b9c46ec\`. Note: W2 verification already flagged this 🔴 critical as unverified and W7 score-clamp downgraded the verdict to 3/5 advisory — so the system itself was skeptical. But the reviewer's *specific suggestion* (503 instead of zero-state) is a real UX improvement regardless of the auth-bypass framing being wrong. Applied that; rejected the framing.

### Applied ✅

1. **503 on upstream-degraded paths** — both catch blocks in \`/api/insights\` now return 503 + a structured \`{ error }\` body instead of \`{ insights: [] }\`:
   - \`fetchUserInstallations\` failure (non-token) → \`503 "GitHub API unavailable, please retry"\`
   - DB-read failure → \`503 "Insights service temporarily unavailable"\`
   
   The dashboard now distinguishes "actually no data yet" (200 + empty) from "upstream is having issues" (503). Better UX signal than the prior silently-fake zero-state.

2. **Client surfaces the 503 error verbatim** — \`InsightsClient\` parses the 503 body's \`error\` field and shows that text in the error panel, instead of "HTTP 503".

### Disagree 🟡

3. **\`:50\` / \`:61\` "auth bypass" framing** (false). The response shape \`{ insights: [] }\` is **information-free**. An attacker cannot distinguish "installation exists / I lack access / I have access but no data" — all three return the same body. No discrimination surface = no probing attack. The 503 fix above DOES close the *operational-information* leak (attacker could distinguish "GitHub API degraded" via status code over time), but that's a separate, real concern from the false "auth bypass" framing the reviewer used.

4. **\`:72\` "info disclosure via error logging"** — the \`console.warn("[/api/insights] ...", err)\` pattern is consistent across every handler in the codebase (lambda/, server/review-processor.ts, and the FB-A/B/C disposition writers). Tightening this one log in isolation creates inconsistency without addressing the broader concern. If this becomes a security priority, the right move is a structured-logger migration across all handlers in one PR.

### Notable

The W2 + W7 chain did its job: the verifier couldn't confirm the critical, the score-clamp downgraded the verdict to advisory, and the rendered review explicitly says "Downgraded to advisory — review carefully, but the PR is not blocked on unverified concerns." Even if I'd disagreed entirely and made no changes, the safety-net behaviour was working. 

The equivalent fall-through in \`/api/settings\` has the same shape — not touching it here to keep the diff scoped, but it should get the same 503 treatment in a follow-up.

Local: \`pnpm run build\` 12/12; \`pnpm run test:coverage\` 20/20.

Issue #182, opened 2026-05-26
FP-F: inline /resolve reply does not persist key, finding re-emits on next review


santthosh-mb:
### Symptom

Posted `resolved` (matching `detectResolveIntent`) as the PR author on a MergeWatch inline thread. On the next synchronize event, the **same critical** was re-emitted as `↻ Still present` in the "📎 Previously reported findings" section — FP-F's stable-key persistence didn't take effect.

### Observable on

[santthosh/mergewatch-fixtures#70](https://github.com/santthosh/mergewatch-fixtures/pull/70) (E2E-35 fixture).

Sequence:

- **19:52 UTC** — round-1 review, inline critical at `src/admin-endpoint.ts:8` ("Unauthenticated admin endpoint exposes sensitive data")
- **23:55:13 UTC** — PR author replied `resolved` to the inline thread via `gh api .../pulls/comments/<id>/replies`
- **23:55:13 UTC** — a `COMMENTED` Review event was submitted (empty body) at the same second — suggests the resolve webhook fired and something handled it
- **23:55:35 UTC** — re-review (after empty `--allow-empty` sync commit) **re-raised the same critical** → 🔴 1/5 CHANGES_REQUESTED

Re-review's rendered "📎 Previously reported findings" collapsible:
```
**↻ Still present (1)**

- `src/admin-endpoint.ts:8` — Unauthenticated admin endpoint exposes sensitive user data
```

Same file, identical code (no commit touched the critical's line), so W9 fingerprint matching identified the carry-over — but FP-F's suppression did not fire.

### Notable: no bot confirmation reply

Compare to FB-D (`/mergewatch reject`) on [PR #75](https://github.com/santthosh/mergewatch-fixtures/pull/75) where the bot posted *"Got it — recording as `style-disagreement`. This finding's pattern won't be re-raised on similar code unless conditions change."* within seconds.

The FP-F path **did not post any acknowledgement reply** to the `resolved` message. This suggests `handleInlineReply` either:
- (a) `detectResolveIntent` didn't match on the plain `"resolved"` body, or
- (b) it matched but `findingMatchKeys` returned `[]` (fail-safe path triggered).

The empty-body `COMMENTED` Review event submitted at the exact second of the resolve reply does suggest *something* fired — but not the user-visible parts of FP-F.

### Hypotheses

1. **`extractInlineCommentTitle`** tripping on this specific inline body shape (`**🔴 <title>**`).
[...]

Issue #181, opened 2026-05-26
FB-A: FindingDispositionRecord writers not persisting (dispositions table empty across dev + prod)


santthosh-mb:
### Symptom

Both `mergewatch-finding-dispositions-dev` and `mergewatch-finding-dispositions-prod` DynamoDB tables are empty (0 items each), despite 19 fixture PRs being reviewed on the fixtures repo over the past several days. Confirmed via `aws dynamodb scan --select COUNT`.

### Corroborating signal

`mergewatch-installation-fp-insights-prod` contains 24 rows (3 windows × 8 installations) but **every row has `totalFindingsSurfaced: 0`, `totalDisputes: 0`, `totalAgreements: 0`, `totalSilentDrops: 0`, `perCategory: {}`, `topClusters: []`** — FB-E nightly rollup is running (generatedAt timestamps match nightly cron) but reading FB-A's empty source.

Sample row from `mergewatch-installation-fp-insights-prod`:

```json
{
  "installationId": "115269590",
  "window": "30d",
  "totalFindingsSurfaced": 0,
  "totalDisputes": 0,
  "totalAgreements": 0,
  "totalSilentDrops": 0,
  "perCategory": {},
  "perRepo": {},
  "topClusters": [],
  "generatedAt": "2026-05-23T03:00:38.904Z"
}
```

### Blast radius

Blocks the entire feedback/dashboard family:

- **E2E-37 FB-A** (storage writers) — direct fail
- **E2E-38 FB-B** (silent-drop counter) — can't increment, no rows to increment against
- **E2E-39 FB-C** (reaction → `disputeCount`/`agreementCount`) — can't increment, no rows
- **E2E-40 FB-D** (`/mergewatch reject`) — bot confirmation reply works ([#75](https://github.com/santthosh/mergewatch-fixtures/pull/75)), but `rejectReasons[]` persistence can't be verified
- **E2E-42–47** dashboard charts — all rendering empty/zero states
- **E2E-48 FB-L** `{{KNOWN_FP_PATTERNS}}` — blocked, no clusters to inject even if shipped
- **E2E-53 FP-J L1** dispute-aware reconcile — `loadCategoryDisputeRates` would return `{}` ⇒ softener never fires

### Hypotheses (ordered by likelihood)

[...]

Issue #183, opened 2026-05-26
Verdict text references a Critical that is missing from the rendered findings table + state/check mismatch (FP-G + FP-L interaction)


santthosh-mb:
### Symptom

On [santthosh/mergewatch-fixtures#71](https://github.com/santthosh/mergewatch-fixtures/pull/71) (E2E-36a fixture, linter-present case):

- **Verdict subtitle reads:** *"2/5 — Needs fixes — Critical security vulnerability (path traversal) and unhandled file system errors present; multiple warnings on code quality; needs fixes before merging."*
- But the rendered "Requires your attention" table contains **only 1 warning + 1 info**, no Critical row.
- There is no `### 🔴 Critical (N)` section in the body.
- There is no `### ⚠️ Unverified concerns` section either.
- **Formal Review state = `CHANGES_REQUESTED`, but `MergeWatch Review` check conclusion = `SUCCESS`** (state / check mismatch).

### Control case (same code, different config)

[PR #72 (E2E-36b, no eslint config)](https://github.com/santthosh/mergewatch-fixtures/pull/72) — **same** `src/style-bait.ts` overlay, but no `eslint.config.mjs` at the repo root. The path-traversal Critical **renders normally** in the table, inline comment, and Critical section:

```
| 🔴 | `src/style-bait.ts:14` | Path traversal vulnerability in file read operation |

### 🔴 Critical (1)
- **`src/style-bait.ts:14`** — Path traversal vulnerability in file read operation
  …
```

So the Critical *was* emitted on #71 but dropped from all three visual surfaces somewhere downstream — likely an interaction between FP-G's linter-aware deferral and FP-L's verification-aware rendering. The state/check mismatch suggests one path (review-event submission) didn't get the memo that the actionable findings count had dropped to 0.

### Hypotheses

1. **FP-L dropped the Critical from `actionFindings` / `buildInlineComments` / the Critical-section header** (because the verifier returned `unverified` only on the linter-present run), but `reviewEventForScore` / `mergeScoreToReviewEvent` wasn't re-evaluated against the post-filter findings → state stays at `CHANGES_REQUESTED`. The check conclusion meanwhile follows the post-filter findings → `SUCCESS`.
2. **The "Unverified concerns (N)" section's render path isn't firing** when N > 0, leaving the dropped Critical with no surface at all (no inline, no action-items row, no Critical section, no Unverified-concerns section).
[...]

PR #185, opened 2026-05-26
fix(fp-f): enrich inline-resolve keys with fingerprint + diagnose drop points

santthosh-mb:
Refs #182.

## What's in the issue

The PR author replied `resolved` on a MergeWatch inline thread. The GraphQL `resolveReviewThread` mutation fired (an empty `COMMENTED` Review event landed at the same second), but on the next sync the same critical re-emitted as **↻ Still present**. No bot acknowledgement, no actionable logs — the operator was left to choose between 4 hypotheses.

## What I could and couldn't verify from code alone

I could not reproduce against live CloudWatch logs from this repo. What I *did* verify by reading the SaaS code path on `upstream/main`:

| Hypothesis from #182 | Verdict from code review |
|---|---|
| 1. `extractInlineCommentTitle` tripping on `**🔴 <title>**` | Regex is single-line non-greedy; title `Unauthenticated admin endpoint exposes sensitive user data` matches cleanly. Not the bug. |
| 2. `detectResolveIntent` regex narrower than RUNBOOK | `affirmative: /^(resolved|please resolve|mergewatch resolve|yes,? resolve)[.!\s]*$/` matches bare `resolved`. Unit-tested at `inline-reply.test.ts:27-32`. Not the bug. |
| 3. `inline_resolved_keys` column missing | Migration `0002_bouncy_sally_floyd.sql` adds it for Postgres; Dynamo is schemaless and `updateStatus` writes the attribute dynamically. Not the bug. |
| 4. Cross-cutting with #181 | Author already noted this only affects the analytics side-effect, not suppression. |

## What this PR actually fixes

Two issues that are real either way and worth fixing on their own merits:

### 1. Title-only persistence is brittle (very plausible root cause for #182)

`deriveResolvedFindingKeys` builds keys via `findingMatchKeys({ file, line: 0, title })` — only the **title** key (`file::T::title`), because the function has no access to the original finding's fingerprint. If the next round's LLM rewords the title even slightly on unchanged code, `partitionDisputed`'s union check misses (no title-key match, and the fingerprint key was never persisted).

Fix: add a small exported helper `enrichResolvedFindingKeys(resolvedKeys, previousFindings)` that joins the resolved seed against the prior review's findings on shared keys, then unions in the full `findingMatchKeys(f)` — which carries `file::F::<fingerprint>` when present. Both handler wrappers (lambda `review-agent.ts` and server `review-processor.ts`) now persist the enriched set, so suppression survives title drift on unchanged code.

### 2. The persistence block is silent on the unhappy paths

The original guard was four ANDed conditions (`latestReview` present, action resolved, resolved keys present, length > 0) producing zero log lines on any miss. That's exactly what made #182 hard to diagnose.

Fix: invert the structure to log *why* a drop happened at each gate:

- `[fp-f] no prior complete review found for <repo>#<pr> — inline-resolve memory not persisted` (covers hypothesis: \`prevReviews\` query returned nothing)
- `[fp-f] resolve fired but no resolved keys derived ... — root inline comment likely missing path or **🔴 <title>** shape` (covers hypothesis 1 + 2 above + the older-comment-shape regression)
- On success: `[fp-f] persisted N inline-resolved key(s) (M after fingerprint enrichment) on <repo>#<pr>` — the M-vs-N gap surfaces whether enrichment fired

So the next time inline-resolve fails to suppress, the CloudWatch trail makes the cause obvious instead of forcing a 4-hypothesis dance.

## Tests

6 new cases for `enrichResolvedFindingKeys` covering: title-drift enrichment, missing-fingerprint passthrough, no-match (independent findings), seed-already-carries-fingerprint, multi-finding parallelism, and dedup. Existing 35 inline-reply tests unchanged.

## Validation

- `pnpm run typecheck` — 20/20 pass
- `pnpm --filter @mergewatch/core test` — 652/652 (41 in inline-reply.test.ts)
- `pnpm --filter @mergewatch/lambda test` — 57/57
- `pnpm --filter @mergewatch/server test` — 79/79

## Test plan (post-merge)

- [ ] Re-run the E2E-35 fixture flow on [mergewatch-fixtures#70](https://github.com/santthosh/mergewatch-fixtures/pull/70):
  1. Apply fixture, wait for round-1 review with inline critical
  2. Reply `resolved` to the inline thread
  3. Empty `--allow-empty` sync commit
  4. Confirm next review *does not* re-raise the critical
  5. Confirm CloudWatch shows `[fp-f] persisted ... (2 after fingerprint enrichment)` then `[fp-f] unioned 2 inline-resolved key(s) into disputedKeys` on the re-review
- [ ] If suppression *still* fails, the new warn lines should now name the gate that dropped — file a follow-up with that log line

### DIFF
--- a/packages/core/src/agents/inline-reply.test.ts
+++ b/packages/core/src/agents/inline-reply.test.ts
@@ -5,6 +5,7 @@ import {
   handleInlineReply,
   detectResolveIntent,
   parseRejectIntent,
+  enrichResolvedFindingKeys,
   REJECT_CATEGORIES,
   MAX_BOT_REPLIES,
 } from './inline-reply.js';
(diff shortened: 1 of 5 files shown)


santthosh-mb:
Pushed `14861fa` addressing the review feedback:

**1. Misleading success log on database failure** (warning + clustered duplication concern)

Extracted the persist+log block into a new `persistInlineResolveMemory` helper in `@mergewatch/core` that uses `try { await; log success } catch { log failure }` — so the success log is gated on actual `await` completion. The old `.catch + always-log` shape that emitted both warn AND success on failure is gone. Regression-tested directly:

```ts
it('reports persisted=false + warn-logs (does NOT log success) when updateStatus throws', ...)
```

**2. Magic number 500** (info)

Extracted as `MAX_INLINE_RESOLVED_KEYS = 500` in core with a comment explaining the rationale (DynamoDB 400 KB item-size limit + parity with the W3 triage path's defensive cap). Used by the helper, exported for symmetry.

**3. Duplicated enrichment + persistence logic** (clustered into #1)

Same helper extraction eliminates the ~30-line duplication between `packages/lambda/src/handlers/review-agent.ts` and `packages/server/src/review-processor.ts`. Each handler shrinks to a single call:

```ts
if (result.action === 'resolved') {
  await persistInlineResolveMemory({
    reviewStore,
    latestReview,
    resolvedFindingKeys: result.resolvedFindingKeys,
    repoFullName,
    prNumber,
  });
  if (result.resolvedFindingKeys?.length) {
    await recordDisputes(dispositionStore, installationId, repoFullName, result.resolvedFindingKeys);
  }
}
```

`recordDisputes` is intentionally outside the helper — its dispute-count signal is decoupled from the `inlineResolvedKeys` persistence (we still want the analytics count even when the persist write fails).

---

**Aside — the "multiple warnings" prose on the rendered review**

The verdict subtitle on this PR reads *"Multiple warnings present regarding error-handling ... code duplication across handlers, and potential DoS vulnerability in array processing"* — but the rendered table shows only **1 warning + 1 info** (with the duplication concern clustered as a W10-related-finding into the success-log warning).

That's the same shape as #183: orchestrator wrote the verdict prose seeing multiple findings, then W10 clustering merged some + the verifier may have dropped one, and `reconcileMergeScore` doesn't regenerate the reason text. PR #186 only handles the *criticals-dropped* case; the *warnings-clustered-or-dropped* case would need a similar narrative-staleness fix. Happy to extend if useful — it would broaden the same orchestrator-vs-post-filter consistency story.

---

**Validation**

- `pnpm --filter @mergewatch/core test` — 658/658 (47 inline-reply: 35 original + 6 enrich + 6 persist)
- `pnpm --filter @mergewatch/lambda test` — 57/57
- `pnpm --filter @mergewatch/server test` — 79/79

PR #187, opened 2026-05-27
fix(fp-l): append narrative-staleness note when post-orchestrator filtering drops findings

santthosh-mb:
Follow-up to PR #186 (which addresses #183) and PR #185 review feedback.

## What's broken

On [PR #185](https://github.com/santthosh/mergewatch.ai/pull/185) the verdict subtitle read:

> *"🟡 3/5 — Review recommended — **Multiple warnings present** regarding error-handling (misleading success logs on database failures), code duplication across handlers, and potential **DoS vulnerability** in array processing — review and fixes recommended before merging."*

But the rendered "Requires your attention" table contained **1 warning + 1 info** — with the duplication concern W10-clustered as a *related concern* under the success-log warning, and the DoS vulnerability nowhere to be found (verifier-dropped). The orchestrator's prose named 3 issues; the rendered table surfaces 1. The reader has no way to know that two of the named issues were dropped or clustered.

PR #186 addresses the closely-related case where dropped *criticals* leave both the score and the prose stale. This PR handles the residual case where the *score* is fine but the *prose* names dropped/clustered findings.

## Fix

Extends `reconcileMergeScore` with an `orchestratorWarningsCount` parameter (paired with the existing `orchestratorCriticalsCount` from #186). On the orchestrator-passthrough branch (where neither the noActionItems / pure-improvement / net-improvement / W7 / FP-J L1 / FP-L #186 branches fire), when the pre-filter action-finding count exceeds the post-filter count, append a single-line clarifying note:

> *"Note: 2 findings were dropped or clustered by post-orchestrator filtering — the rendered list is authoritative."*

What it preserves:
- The orchestrator's *specific narrative* (it's often valuable when it's accurate)
- The score (we're not clamping; this is purely a prose-staleness fix)
- All earlier branches' deterministic reasons (they're unaffected — the append only fires on the orchestrator-passthrough fall-through)

What it adds:
- A factual note when the post-filter set is meaningfully smaller than what the orchestrator saw
- Idempotent guard: a marker-phrase check prevents a re-reconcile from stacking duplicate notes
- Full back-compat: callers that omit both counts get the verbatim orchestrator reason (pre-fix behavior)

The pipeline caller (`runReviewPipeline`) now computes both counts from `orchestratorResult.findings` and passes them through.

## What the rendered verdict will look like

For the PR #185 case (orchestrator emitted 3 warnings, post-filter has 1):

> *"🟡 3/5 — Review recommended — Multiple warnings present regarding error-handling (misleading success logs on database failures), code duplication across handlers, and potential DoS vulnerability in array processing — review and fixes recommended before merging.*
>
> *Note: 2 findings were dropped or clustered by post-orchestrator filtering — the rendered list is authoritative."*

For the no-shrinkage case (everything the orchestrator named still renders): nothing changes.

## Tests

7 new cases in the `FP-L narrative-staleness append` describe block:
- PR #185 scenario (orch=3, post=1, plural grammar)
- Singular grammar (delta=1)
- Mixed critical + warning counting
- No-op when nothing was dropped
- Back-compat without counts
- Idempotent guard against double-appending
- Earlier-branch precedence (noActionItems wins, no append)

## Validation

- `pnpm run typecheck` — 20/20 pass
- `pnpm --filter @mergewatch/core test` — 652/652
- `pnpm --filter @mergewatch/lambda test` — 57/57
- `pnpm --filter @mergewatch/server test` — 79/79

## Test plan (post-merge)

- [ ] Inspect a few recent reviews where W10 clustering or verifier drops likely fired — confirm the note appears on the ones with actual shrinkage and doesn't appear on the clean ones
- [ ] Confirm reviews with no orchestrator dropping (score=5 noActionItems, pure-security-improvement, etc.) don't pick up a spurious note

## Sequencing note

This PR is independent of #186 but reads more naturally if reviewed after it (the `orchestratorCriticalsCount` parameter shape was introduced there). If #186 lands first, this becomes a small extension. If this lands first, #186 just needs a trivial rebase to add `orchestratorWarningsCount` to its own input destructuring (already present here).

### DIFF
--- a/packages/core/src/agents/reviewer.test.ts
+++ b/packages/core/src/agents/reviewer.test.ts
@@ -2522,6 +2522,144 @@ describe('reconcileMergeScore', () => {
     });
   });
 
+  // ─── FP-L narrative-staleness fix (follow-up to PR #185 review) ────────
+  //
+  // Scenario: orchestrator emits a richer set of findings than the
+  // post-filter pipeline keeps (verifier drops, W10 clustering, line-
+  // filter, W3 dispute). The orchestrator's narrative text references
+  // findings that no longer render — "Multiple warnings present
+  // regarding X, Y, and Z" when only X is in the table. The reconciled
+  // reason appends a clarifying note so the reader knows the rendered
+  // list is authoritative.
+  describe('FP-L narrative-staleness append', () => {
+    it('appends a clarifying note when orchestrator pre-filter action count exceeds post-filter count', () => {
+      // The PR #185 scenario: orchestrator emitted 3 warnings; W10
+      // clustered one + the verifier dropped one; rendered = 1 warning.
+      const r = reconcileMergeScore({
+        filteredFindings: [warning()],
+        previousFindings: undefined,
+        orchestratorScore: 3,
+        orchestratorReason: 'Multiple warnings present regarding error handling, code duplication, and DoS — review recommended before merging.',
+        orchestratorWarningsCount: 3,
+        orchestratorCriticalsCount: 0,
+      });
+      expect(r.mergeScore).toBe(3);
+      // Original narrative preserved.
+      expect(r.mergeScoreReason).toContain('Multiple warnings present');
+      // Clarifying note appended.
+      expect(r.mergeScoreReason).toMatch(/2 findings were dropped or clustered/);
+      expect(r.mergeScoreReason).toMatch(/rendered list is authoritative/);
+    });
+
+    it('uses singular grammar when exactly one finding was dropped', () => {
+      const r = reconcileMergeScore({
+        filteredFindings: [warning(), warning()],
+        previousFindings: undefined,
+        orchestratorScore: 3,
+        orchestratorReason: 'Two warnings plus a DoS concern.',
+        orchestratorWarningsCount: 3,
+        orchestratorCriticalsCount: 0,
+      });
+      expect(r.mergeScoreReason).toMatch(/1 finding was dropped or clustered/);
+    });
+
+    it('counts criticals + warnings together for the staleness threshold', () => {
+      // Orchestrator: 1 critical + 2 warnings = 3 action findings.
+      // Post-filter: 1 critical + 1 warning = 2.  Delta = 1.
+      const r = reconcileMergeScore({
+        filteredFindings: [critical({ verification: 'verified' }), warning()],
+        previousFindings: undefined,
+        orchestratorScore: 2,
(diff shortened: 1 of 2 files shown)


santthosh-mb:
Force-pushed `2fbc96e` — rebased against main (PR #186 is now in) and addressed the MergeWatch review feedback in the same commit.

**Rebase**

PR #186 added `orchestratorCriticalsCount` to `reconcileMergeScore` first; my branch added both criticals + warnings counts. Resolved by:
- Keeping the existing `orchestratorCriticalsCount` from #186 (powers the verifier-dropped clamp)
- Adding `orchestratorWarningsCount` alongside (powers the narrative-staleness note in this PR)
- Consolidating the two JSDoc blocks into a single shared comment that describes both uses — also addresses the *"verbose / overlapping JSDoc"* info finding
- Keeping both `describe()` test blocks (FP-L #183 verifier-dropped + FP-L narrative-staleness append) side-by-side — they test independent concerns

**Review feedback addressed**

1. ⚠️ **Partial-count provision could trigger spurious notes** — Implemented the suggested guard verbatim: `reconcileNarrativeStaleness` now requires *both* counts together and treats partial provision (only criticals or only warnings) as absent. Added a regression test `partial counts (only criticals OR only warnings) are conservatively treated as absent` that locks the contract.

2. 🧹 **Extract staleness detection logic** (clustered related concern) — Pulled the comparison + dispatch out of the inline conditional and into a `reconcileNarrativeStaleness(reason, criticalsCount, warningsCount, postFilterActionCount)` helper. The caller is now a single named call instead of an inline ternary. The lower-level `appendStalenessNote` retains the actual string concatenation.

3. 🔡 **Magic string** (clustered related concern) — Extracted `STALENESS_NOTE_MARKER = 'rendered list is authoritative'` as a module-level constant. Both the append (note construction) and the idempotency guard (`reason.includes(STALENESS_NOTE_MARKER)`) read from the same source, so the marker can't drift between writer and detector.

4. ℹ️ **Verbose / overlapping JSDoc** — Replaced the two long JSDoc blocks with a single shared comment block over both parameters that explains the two distinct uses (criticals → #183 clamp, sum → #185 follow-up note) in fewer lines.

**Validation**

- `pnpm --filter @mergewatch/core test` — 674/674 (8 FP-L narrative-staleness cases including the new partial-counts guard)
- `pnpm --filter @mergewatch/lambda test` — 61/61
- `pnpm --filter @mergewatch/server test` — 79/79

PR #202, opened 2026-06-20
fix(deps): resolve 15 Dependabot security alerts

santthosh:
## Resolves all 15 open Dependabot alerts

(1 critical · 3 high · 7 moderate · 1 low — every open alert on the default branch.) Code scanning and secret scanning are not enabled, so Dependabot was the only source.

All fixes are version bumps of vulnerable **transitive** deps via the existing `pnpm.overrides` block, plus one direct `vite` pin. No source/runtime behavior changes.

| Severity | Package | From | To | Advisory |
|---|---|---|---|---|
| 🔴 critical | shell-quote | 1.8.3 | 1.8.4 | `quote()` doesn't escape newlines in object `.op` |
| 🟠 high | hono | 4.12.18 | 4.12.25 | (+ 7 moderate hono alerts in the same line) |
| 🟠 high | vite | 8.0.13 | 8.0.16 | dev-server arbitrary file read |
| 🟡 moderate | qs | 6.15.1 | 6.15.2 | `stringify` DoS |
| 🟡 moderate | uuid | 8.3.2 / 10.x | 11.1.1 | missing buffer bounds check (v3/v5/v6) |
| ⚪ low | esbuild | 0.28.0 | 0.28.1 | dev-server arbitrary file read |

### Notes
- **vite** is a dev-only dependency (vitest's test runner — never shipped in the Lambda or dashboard artifacts). The bare `pnpm` override wouldn't dislodge vitest's cached peer resolution, so it's pinned as a direct `devDependency` at `8.0.16` instead.
- **esbuild** override is scoped to the vulnerable `>=0.27.3 <0.28.1` range, so the safe `0.25.x` instances (the Lambda's direct dep) are left untouched.
- **uuid** is a major bump (8/10 → 11) for transitive consumers; gated by the full build + test suite below.

### Verification
- `pnpm run build` — 12/12 ✅
- `pnpm run typecheck` — 20/20 ✅
- `pnpm run test` — all 20 suites green (core 715, server 84, lambda 61, dynamo 24, …) ✅
- Lockfile re-scanned: no vulnerable versions remain (the only `qs@6.15.1` left is `@types/qs`, type-defs only).


santthosh:
@mergewatch review

PR #228, opened 2026-06-25
fix(fb-c): sweep inline reactions on PR close so end-of-PR reactions count (#189)

### LINKED ISSUE #189: FB-C: emoji reactions on inline bot comments do not increment agreementCount / disputeCount
### Symptom

Per [RUNBOOK E2E-39 (FB-C)](https://github.com/santthosh/mergewatch.ai/blob/main/e2e/RUNBOOK.md#e2e-39-fb-c--inline-comment--reactions--disputes), reactions on a `mergewatch[bot]` inline finding comment should map as:

| Reaction | Counter |
|---|---|
| 👎 (`-1`) / 🤔 (`confused`) | `disputeCount` |
| 👍 (`+1`) / ❤️ (`heart`) / 🚀 (`rocket`) | `agreementCount` |

Tested on [santthosh/mergewatch-fixtures#118](https://github.com/santthosh/mergewatch-fixtures/pull/118):
1. Added 👎 to the inline bot comment (id `3308330603`) — expected `disputeCount` += 1
2. Added 🚀 — expected `agreementCount` += 1
3. Removed the 👎 (monotonic check) — `disputeCount` should stay incremented

### Actual

Aggregate FB-A counters across all 34 disposition rows after this run:

```
disputeCount      :  3   ← fully accounted for by W3 triage (#101) + /mergewatch reject (#119); no contribution from the 👎
agreementCount    :  0   ← 🚀 not counted on any row
```

The admin-endpoint finding row that #118's inline comment targeted (`src/admin-endpoint.ts::T::Unauthenticated admin endpoint exposes user data`) has:
- `surfaceCount: 3` (correctly counts the 3 surfacings across #114, #118, #119)
- `disputeCount: 1` (from `/mergewatch reject` on #119 only — `rejectReasons[]` timestamp `2026-05-27T04:20:20.305Z` matches the reject call exactly)
- `agreementCount: 0`

The 👎 added on #118 (~04:20:10) and 🚀 added a few seconds later left no trace.

### Hypotheses

1. **The reactions webhook handler isn't wired** — issue/comment-reaction events on inline review comments aren't being routed to FB-C's increment path.
2. **Handler IS wired but filters on the wrong predicate** — e.g. requires `comment.path` to be present, or rejects reactions on `IN_REPLY_TO`-style comments, or requires the `INLINE_BOT_COMMENT_MARKER` check to read the body in a way that doesn't match.
3. **Counter increment is best-effort and silently swallows write errors** (like the FB-A symptom we just fixed in #181) — if so, check IAM `dynamodb:UpdateItem` on the dispositions table.

### Cross-check that proves the rest of FB-A works

- W3 triage (top-level PR comment on #101) → `disputeCount: 1` on the matching `swallow.ts` row ✅
- `/mergewatch reject` (inline slash command on #119) → `disputeCount: 1` + populated `rejectReasons[]` on the matching admin-endpoint row ✅
- Only the **emoji-reaction** path is silent.

### Refs

- RUNBOOK [E2E-39 (FB-C)](https://github.com/santthosh/mergewatch.ai/blob/main/e2e/RUNBOOK.md#e2e-39-fb-c--inline-comment--reactions--disputes)
- Related: #181 (FB-A storage — fixed)
- Test PR: santthosh/mergewatch-fixtures#118

### Repro

```bash
# Apply the fixture
cd mergewatch-fixtures && ./scripts/apply-fixture.sh 39-inline-reactions
# Wait for review.  Then add reactions:
gh api -X POST repos/santthosh/mergewatch-fixtures/pulls/comments/<inline-id>/reactions -f content='rocket'
gh api -X POST repos/santthosh/mergewatch-fixtures/pulls/comments/<inline-id>/reactions -f content='-1'
# Check the matching disposition row's agreementCount + disputeCount.
# Expected: both incremented.  Actual: both remain 0 / unchanged.
```

santthosh:
**Closes #189.**

## Root cause (verified, not the filed hypotheses)
FB-C *is* implemented and wired, and `listReviewComments` *does* return the `reactions` summary. The real gap is the **trigger**: `pollAndRecordInlineReactions` only runs **during a review**, and GitHub emits **no webhook for reactions** — so a reaction added after the *final* review is never polled.

Ground truth on `santthosh/mergewatch-fixtures#118`:
```
bot inline comment 3308330603 → reactions: { rocket: 1 }   (still there)
#118 reviews: ONE review at 03:56  — the 🚀 was added ~04:20, AFTER it
```
No review ran after the reaction → it was never counted (`agreementCount: 0`). That's the common case: people react *when they read the review*.

## Fix
Capture reactions one final time on the terminal `pull_request` **`closed`** event (which we already handle for TTM lifecycle). New core helper **`sweepInlineReactionsOnClose`**:
- loads the latest review's reaction snapshot,
- runs the **same** poll (so only reactions added *since the last poll* are counted — no double-count with in-review polls),
- persists the refreshed snapshot so a rare *re-open → react → re-close* can't re-count.

Wired into **both** runtimes' close handlers — lambda `webhook.ts` (module review/disposition stores + `authProvider` octokit) and server `webhook-handler.ts` (`deps.reviewStore` / `deps.dispositionStore` / `deps.authProvider`). Best-effort; never blocks the close path. Net: a reaction is captured on the **next review OR when the PR closes**, whichever comes first.

## Tests
`disposition-writer.test`: sweep counts reactions added since the last review **and persists the refreshed snapshot**; captures on a first-ever poll (no prior review); no-ops (no API call) without a disposition store. RUNBOOK **E2E-39** now documents the poll/close-sweep capture timing + the #189 failure mode. `build`+`typecheck`+`test` 20/20 (core **770**).

### DIFF
--- a/e2e/RUNBOOK.md
+++ b/e2e/RUNBOOK.md
@@ -1574,23 +1574,27 @@ Branch: `fixture/38-quiet-drop`. A PR with a finding that the orchestrator's con
 
 Reaction *removal* is a no-op (signal stays monotonic). Anonymous: we count, we don't store reactor identity.
 
+**Capture timing (#189):** GitHub does NOT emit a webhook for reactions, so MergeWatch **polls** them — folding a single `listReviewComments` call (the per-comment `reactions` summary) into the post-pipeline path and counting only the positive delta vs the snapshot persisted on the prior review. Because a reaction is usually added *after* the final review (people react when they read it), an in-review poll alone would miss it — so a final sweep **also runs on the terminal `closed` event** (`sweepInlineReactionsOnClose`, both runtimes). Net: a reaction is captured on the **next review of the PR OR when the PR closes**, whichever comes first.
+
 **Setup**
 
 Branch: `fixture/39-inline-reactions`. A PR with at least one inline-comment-eligible finding:
 1. Confirm `FindingDispositionRecord` row exists post-review with `disputeCount = 0`, `agreementCount = 0`.
-2. Add 👎 to the inline bot comment → confirm `disputeCount = 1`.
-3. Add 🚀 → confirm `agreementCount = 1`.
-4. Remove the 👎 → confirm `disputeCount` stays at 1 (monotonic).
+2. Add 👎 to the inline bot comment, then **trigger a poll** — push a commit (re-review) or **close the PR** — and confirm `disputeCount = 1`.
+3. Add 🚀, trigger another poll → confirm `agreementCount = 1`.
+4. Remove the 👎 before the next poll → after a poll, confirm `disputeCount` stays at 1 (monotonic).
 
 **Expected outcomes**
 
 - [ ] 👎 / 🤔 ↔ `disputeCount` mapping fires per-reaction
 - [ ] 👍 / ❤️ / 🚀 ↔ `agreementCount` mapping fires per-reaction
+- [ ] **#189** — a reaction added *after the final review* is still captured when the PR is **closed** (the close-sweep), not silently lost
 - [ ] Reactions on the TOP-level bot comment continue to populate `ReviewItem.reactions` separately (back-compat)
 - [ ] Reactions added by `mergewatch[bot]` itself are ignored (no self-counting)
 
 **Failure modes**
 - ❌ Reaction removal decrements the counter (must be monotonic)
+- ❌ **#189** — a reaction added after the last review never increments because no poll ever ran (must be captured by the close-sweep)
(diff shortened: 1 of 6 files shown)


santthosh:
Addressed the substantive review finding (snapshot double-counting → now merges all prior snapshots, order-independent). The two residual warnings are false positives, justified:
- **`prNumber` in `queryByPR`** — it's a typed integer from GitHub's payload and is passed as a *bound* query parameter (`:pr` via `begins_with`), not string-concatenated, so it isn't injectable. Same call shape the review-agent already uses.
- **property access on `target`** — `target` is a guarded, typed `ReviewItem` (`completes[0]`, inside `if (target && …)`); `prNumberCommitSha` is its always-present sort key. Identical to the existing `persistInlineResolveMemory` pattern.

Merging.

Issue #231, opened 2026-06-25
Self-hosted: show per-PR + dashboard cost when model pricing is set (don't silently hide unpriced)


santthosh:
## Problem

On self-hosted deployments, the **per-PR cost** (the "Est. cost" line in the review comment) and the **Cost & Impact** dashboard block are blank/zero, even though tokens and other Impact metrics (cycle time, engagement) show fine.

### Root cause — it's the pricing table, not a deployment-mode gate

There is **no explicit suppression** of cost in self-hosted. Both runtimes pass `estimatedCostUsd` into the comment and cost stores identically:
- Self-hosted: `packages/server/src/review-processor.ts:660` (comment), `:814` (cost record)
- SaaS: `packages/lambda/src/handlers/review-agent.ts:792` (comment), `:949` (cost record)

The cost is hidden because **`estimateCost()` returns `null` for any model not in the pricing table** (`packages/core/src/llm/pricing.ts`). `DEFAULT_PRICING` only covers Bedrock Anthropic IDs and a handful of direct-Anthropic IDs. Self-hosted users typically run:
- Ollama (`llama3`, …), LiteLLM / OpenAI-compatible (`gpt-4o`, …), or
- a newer/differently-named Anthropic ID not yet in the table.

When the model is unpriced:
1. `estimateCost()` → `null` → `result.estimatedCostUsd` is `null`.
2. The review-comment "Est. cost" line is skipped (`comment-formatter.ts:503` gates on `estimatedCostUsd > 0`).
3. The cost row is denormalized with `cost_usd = null` ("unpriced"); the rollup (`core/src/insights/cost.ts`) excludes it from money → `totalCostUsd = 0`, `avgCostPerReview = null`, `unpricedReviewCount = N`.
4. Dashboard Cost section shows `$0.00` / `—` (or is effectively empty).

This is by design (we never want a mis-priced model to silently pull averages toward 0) — but the result is that self-hosted operators see no cost at all out of the box.

## Desired behavior

**Start showing per-PR + dashboard cost whenever model pricing is known** — i.e. make the "priced model → cost appears" path work out of the box and be discoverable, rather than requiring every self-hosted operator to hand-author a pricing override before they see anything.

[...]

PR #234, opened 2026-06-25
feat(cost): global LLM_MODEL_*_PRICE env vars for self-hosted pricing (#233)

### LINKED ISSUE #233: Self-hosted: global LLM_MODEL_*_PRICE env vars to price the LLM_MODEL (esp. Bedrock inference-profile ARNs)
## Problem

Self-hosted operators commonly set the review model **globally** via the `LLM_MODEL` environment variable. When that value is a model MergeWatch can't price by ID — most notably a **Bedrock application inference profile ARN** (e.g. `arn:aws:bedrock:us-west-2:…:application-inference-profile/hwswp7wpd6c5`), but also any custom LiteLLM/Ollama ID — every review records `cost_usd = null` and the dashboard shows `$0` / "unpriced".

Today the only no-IAM way to price such a model is a **`pricing:` block in `.mergewatch.yml`** (added in #231). But that's **per-repo**, while `LLM_MODEL` is **global**. So an operator who sets one env var to configure the model for every repo would have to copy the same `pricing:` block into every repo's `.mergewatch.yml`. The config surfaces don't match.

## Context — how the model is set globally today

`LLM_MODEL` overrides both `model` and `lightModel` for every review, beating `.mergewatch.yml`, installation config, and dashboard settings (`packages/server/src/review-processor.ts:371-379`):

```js
// YAML provides base, dashboard settings override, env var model overrides all.
const modelOverride = process.env.LLM_MODEL;
const config = mergeConfig({
  ...(yamlConfig ?? {}),
  ...(installation?.config || {}),
  ...settingsOverrides,
  ...(modelOverride ? { model: modelOverride, lightModel: modelOverride } : {}),
});
```

Cost is keyed by the recorded model ID (the `LLM_MODEL` value), looked up in `DEFAULT_PRICING` / `customPricing` via `estimateCost`. There is no global pricing knob to pair with the global `LLM_MODEL`.

## Proposal

Add **global pricing env vars** (self-hosted server) that register a price for whatever `LLM_MODEL` is set to — mirroring how the model itself is configured. Zero IAM, zero per-repo config, no Bedrock API call.

```bash
LLM_MODEL=arn:aws:bedrock:us-west-2:029800430051:application-inference-profile/hwswp7wpd6c5
LLM_MODEL_INPUT_PRICE_PER_1M=5     # USD per 1M input tokens
LLM_MODEL_OUTPUT_PRICE_PER_1M=25   # USD per 1M output tokens
```

Because `LLM_MODEL` already sets `model` **and** `lightModel` to the same value, a single input/output pair prices the whole review. The env-derived price becomes a `customPricing` entry keyed by the `LLM_MODEL` value and is merged into the `customPricing` already passed to the pipeline (`review-processor.ts:569`).

## Design details

- **Where:** server only (`LLM_MODEL` is a server-path override; SaaS/Lambda uses system inference profiles that are already priced by ID). Read the env once, build `{ [LLM_MODEL]: { inputPer1M, outputPer1M } }`.
- **Threading:** merge into the `customPricing` passed to `runReviewPipeline` (and any other path that passes `customPricing`, e.g. the inline-reply path that also honors `LLM_MODEL`).
- **Precedence:** per-repo `.mergewatch.yml` `pricing:` (most specific) overrides the global env price for the same model ID; env price is the global default. Document this.
- **Validation:** both vars must be finite, non-negative numbers; if only one is set, ignore (with a one-time warn) rather than half-pricing. `0`/`0` is valid → records a real priced $0 (local model).
- **No effect without `LLM_MODEL`:** the price vars are scoped to the `LLM_MODEL` value; if `LLM_MODEL` is unset they're a no-op (a future `LLM_PRICING` JSON map could generalize to multiple models — out of scope here).
- **Interplay with #231:** reuses the same `customPricing` plumbing and the dashboard "unpriced" hint; an operator who sets these vars stops seeing the hint.

## Acceptance criteria

- [ ] Setting `LLM_MODEL` + `LLM_MODEL_INPUT_PRICE_PER_1M` + `LLM_MODEL_OUTPUT_PRICE_PER_1M` makes per-PR `Est. cost` and the dashboard Cost block show real spend for that model — including an opaque inference-profile ARN — with no `.mergewatch.yml` change.
- [ ] A per-repo `.mergewatch.yml` `pricing:` entry for the same model overrides the env price.
- [ ] `0`/`0` records a priced $0 (not "unpriced"); a single var set (missing its pair) is ignored with a one-time warn; non-numeric/negative values are ignored.
- [ ] No effect when `LLM_MODEL` is unset; SaaS/Bedrock behavior unchanged.
- [ ] Documented in `.env.example` + README self-hosted config; unit tests cover parse, merge/precedence, and validation.

## Alternatives considered

- **Per-repo `.mergewatch.yml` `pricing:`** (#231) — works today but doesn't match a globally-configured model (needs the block in every repo).
- **Auto-resolve the inference profile** via Bedrock `GetInferenceProfile` → price the base model (zero config). Cleaner for ARN users but requires a new IAM read permission (`bedrock:GetInferenceProfile`) and is Bedrock-specific. Could be a separate follow-up; this env-pricing ticket is provider-agnostic and IAM-free.

## Affected files

- `packages/server/src/review-processor.ts` (read + merge env pricing into `customPricing`; both review + inline-reply paths)
- possibly a small helper in `@mergewatch/core` (`packages/core/src/llm/pricing.ts`) for parse/validate, to keep it testable
- `.env.example`, `README.md` (self-hosted config)

santthosh:
## Summary

Closes #233. Self-hosted deployments set the review model globally via `LLM_MODEL` (it overrides `model` + `lightModel` for every repo). But the only no-IAM way to price an unknown model — especially a **Bedrock application-inference-profile ARN** like `arn:aws:bedrock:…:application-inference-profile/…` — was a *per-repo* `.mergewatch.yml` `pricing:` block (#231). That's a config-surface mismatch with a globally-configured model. This adds a global pricing knob that pairs with `LLM_MODEL`.

## Changes

- **core — `parseEnvModelPricing()`** (pure, env-agnostic): builds a single-entry `customPricing` map from raw `LLM_MODEL` + price strings. Validates both prices are present, finite, non-negative; allows `0/0`; rejects blank/non-numeric/negative. Exported from `@mergewatch/core`.
- **server — full review path:** read `LLM_MODEL_INPUT_PRICE_PER_1M` / `LLM_MODEL_OUTPUT_PRICE_PER_1M`, merge as `{ ...env, ...config.pricing }` so a per-repo `pricing:` still **wins**, pass as `customPricing`. One-time `[cost]` warn when `LLM_MODEL` is set but the price vars are partial/invalid (so a typo doesn't silently read as $0).
- **server — inline-reply path:** thread `customPricing` through `handleInlineReply`. This also **closes a pre-existing gap**: inline replies called `estimateTotalCost()` with no custom pricing, so they were always unpriced for custom models.
- **docs:** `.env.example` + README (precedence + `0/0` convention); RUNBOOK **E2E-67**.

## Behavior

```bash
LLM_MODEL=arn:aws:bedrock:us-west-2:…:application-inference-profile/abc123
LLM_MODEL_INPUT_PRICE_PER_1M=5
LLM_MODEL_OUTPUT_PRICE_PER_1M=25
```
→ per-PR `Est. cost` and dashboard Cost show real spend, for both full reviews and inline replies, with **no** `.mergewatch.yml` change and **no** IAM. `0/0` = priced $0 (local model). Per-repo `pricing:` overrides the env price. Partial/invalid → one-time warn, ignored.

## Acceptance criteria (from #233)
- [x] `LLM_MODEL_*_PRICE_PER_1M` prices the `LLM_MODEL` (incl. an ARN) with no per-repo config
- [x] Per-repo `.mergewatch.yml` `pricing:` overrides the env price
- [x] `0/0` → priced $0; single/blank/non-numeric/negative → ignored with one-time warn
- [x] No effect when `LLM_MODEL` unset; SaaS/Bedrock unchanged
- [x] Documented (`.env.example` + README); unit tests for parse, merge/precedence, validation, and the inline-reply path

## Tests
New unit tests: `parseEnvModelPricing` (valid / decimal / `0,0` / missing / blank / non-finite / negative / feeds `estimateCost`) and the inline-reply cost path (unpriced ARN → null; priced via `customPricing` → spend). `build` + `typecheck` + `test` green workspace-wide. No schema change.


santthosh:
The remaining 4/5 finding is a false positive (the reviewer's own summary calls it speculative / not a demonstrated defect): the blank-string guard rejects whitespace-only values *before* `Number()` is reached, and the latest commit additionally trims before coercion with symmetric tests. No real defect; merging.

Issue #261, opened 2026-08-12
OSS Program: honor free hosted access for approved open-source projects (sponsored-review entitlement)


santthosh:
## Context

[mergewatch.ai/open-source](https://mergewatch.ai/open-source) promises approved open-source projects **free hosted MergeWatch access** — public repos only, manual approval, fair-use limits, no SLA. The docs repeat it at `docs-site/saas/billing.mdx` ("Open-source maintainers … free frontier-model review … beyond the standard free tier"). Applications come in through a live Google Form.

**Nothing in the runtime honors that promise.** The billing gate has no concept of a sponsored install:

- `billingCheck()` (`packages/billing/src/billing-check.ts:23`) is a two-branch decision — `freeReviewsUsed < 5` → allow, else `balanceCents >= 5¢` → allow, else block.
- It's called once at `packages/lambda/src/handlers/review-agent.ts:450`. A block posts an `action_required` check run **and files a GitHub Issue on the repository** (`packages/billing/src/block-notify.ts`).

So the only way to service an approved application today is hand-editing `balanceCents` via `updateBillingFields`. That's the wrong primitive:

1. **It drains.** The program dies quietly when a granted balance hits zero — and the failure mode is MergeWatch filing a *"reviews paused — credits required"* issue on a public OSS repo. That is the single worst outcome this program could produce.
2. **It books sponsorship as revenue.** `totalBilledCents` and the analytics built on it stop meaning anything.
3. **There is no ceiling.** The page promises "fair-use limits"; a raw balance grant has none.
4. **The maintainer sees a dollar balance**, so a sponsored user looking at the billing page reasonably assumes they're about to be charged.

---

## Evaluated and rejected: Stripe coupons / promotion codes

Worth recording why, since "just use a Stripe coupon" is the obvious first instinct.

Stripe coupons and promotion codes discount **invoices, subscriptions, and Checkout Sessions in `payment`/`subscription` mode**. MergeWatch bills through none of those:

- There are **no subscriptions and no invoices** anywhere in the codebase (`grep` for `subscription|invoice|coupon|discount` across `packages/billing`, `packages/lambda`, and the dashboard billing API returns nothing).
- Top-ups are a **direct off-session `paymentIntents.create`** (`packages/billing/src/checkout.ts:106`), not a Checkout Session. Coupons cannot be applied to a raw PaymentIntent at all.
[...]

Issue #264, opened 2026-08-13
Model selection is not wired: prod pinned to Sonnet 4 by a retained CFN parameter, and .mergewatch.yml `model:` is never read


santthosh:
Two independent defects mean **the model MergeWatch actually reviews with is not the model any config file says**, and there is currently no working way to override it per repo.

Found while verifying the baseline for #262 — the MergeWatch review on #263 reported its own model as `claude-sonnet-4`, which matches nothing in the repo.

## Defect 1 — prod runs Sonnet 4; the SAM template default has never applied

```
$ aws lambda get-function-configuration --function-name mergewatch-review-agent-prod \
    --query Environment.Variables.DEFAULT_BEDROCK_MODEL_ID
"us.anthropic.claude-sonnet-4-20250514-v1:0"

$ aws cloudformation describe-stacks --stack-name mergewatch \
    --query "Stacks[0].Parameters[?ParameterKey=='DefaultBedrockModelId']"
[{ "ParameterKey": "DefaultBedrockModelId",
   "ParameterValue": "us.anthropic.claude-sonnet-4-20250514-v1:0" }]
```

Meanwhile `infra/template.yaml:55` declares:

```yaml
DefaultBedrockModelId:
  Type: String
  Default: us.anthropic.claude-opus-4-6-v1
```

**A CloudFormation parameter `Default:` applies only at stack *creation*.** On every update CFN retains the previously-supplied value unless the deploy explicitly overrides it. The stack was created (or once deployed) with the Sonnet 4 value, so it has been pinned ever since — and editing `Default:` in the template has been a silent no-op for prod this whole time.

The Lambda's `LastModified` is `2026-08-11`, so this is not a stale deploy. Current code, retained parameter.

**Fix:** pass the parameter explicitly on deploy (`--parameter-overrides DefaultBedrockModelId=…` in `scripts/deploy.sh`), so the template and the stack can't drift again. Consider dropping the misleading `Default:` or adding a comment that it only binds at creation.

## Defect 2 — the documented per-repo `model:` key is never read

`docs-site/configuration/mergewatch-yml.mdx:124` documents it as functional:

> | `model` | `string` | `us.anthropic.claude-opus-4-6-v1` | Bedrock model ID used for all agents. Any model available in your Bedrock account can be specified. |

[...]

PR #265, opened 2026-08-13
feat(billing): wire OSS repo context through the webhook to the gate (#261, stage 2)

santthosh:
**Stage 2 of 4** for #261. **Stacks on #263** (stage 1) — review that first; this PR's base is `feat/oss-program-stage1-core-entitlement`.

Stage 1 shipped the entitlement dark. This connects it to real reviews.

## What this ships

- `repository.id` and `repository.private` now ride `ReviewJobPayload` as `repoId` / `isPublic`. Neither was forwarded before; they live on the same payload object, so it's one change.
- `review-agent` turns them into the optional repo context and passes it to **both** `billingCheck` and `recordReview`, so the gate and the recorder can't disagree about whether a review was sponsored.
- Block copy is OSS-aware: `billingCheck` now surfaces `ossReason`, and a grant that expired or hit its cap gets a check run + issue pointing at renewal / BYOK / self-hosting instead of "add a credit card".
- `E2E-82` (sponsored review lifecycle) and `E2E-83` (operator grant lifecycle) added to the runbook, plus `docs/pending/oss-program.md`.

## Worth reviewing

**All four enqueue sites, not just `pull_request`.** An `@mergewatch` mention, an inline reply, and a re-requested check each run the same billing gate. Omitting repo context on those paths would silently bill a maintainer we promised free review — so they all carry it, via a shared `ossRepoFields()` helper.

**Mirrored into the self-hosted handler.** Those fields are inert there (billing is behind `isSaas()`), but a shared payload type that means different things depending on which runtime populated it is a trap for the next reader.

**Why `ossReason` exists.** The gate already knew *why* OSS didn't apply; it was only logging it. Surfacing it lets the block path distinguish "this repo was never sponsored" (standard credits copy) from "your grant lapsed / hit its cap" (renewal copy). Only `grant_expired` and `cap_exceeded` switch the copy — `isLapsedOssGrant()` encodes that.

## Note on two changed stage-1 tests

Adding `ossReason` broke two exact-shape `toEqual` assertions from stage 1. They were **tightened, not loosened** — they now assert the new field explicitly (`ossReason: 'grant_expired'` / `'cap_exceeded'`) rather than switching to partial matching.

## Verification

Build ✅ typecheck ✅ full suite ✅ — 123 billing, 64 lambda, 95 server, 818 core.

## Risk

Still effectively dark: nothing writes grant fields until stage 3, so `evaluateOssGrant` returns `no_grant` for every installation and every path behaves exactly as it does today. The one live change is that four job payloads carry two extra fields.

## Deferred

- **Stage 3** — `scripts/grant-oss.sh`.
- **Stage 4** — dashboard OSS state + docs; graduates `docs/pending/oss-program.md`.

### DIFF
--- a/docs-site/saas/billing.mdx
+++ b/docs-site/saas/billing.mdx
@@ -21,6 +21,20 @@ MergeWatch offers **free frontier-model review to open-source maintainers**, bey
 Applications are reviewed individually — tell us about the project and how you'd use MergeWatch.
 </Card>
 
+### How an approved grant behaves
+
+Once a project is approved, reviews on the covered repositories are **sponsored**: no charge, and they don't consume your free reviews or credit balance. Your **Dashboard → Billing** page shows an Open Source Program panel with the covered repositories, what's been sponsored this month, and your fair-use headroom.
+
+Specifics worth knowing:
+
+- **Only the repositories named in the grant are covered.** Approval is per project, not per organization — other repositories in the same installation bill normally. Ask us to add a repository and we will.
+- **Coverage requires the repository to be public at review time.** If a covered repository is made private, its next review is billed normally rather than sponsored.
+- **Renaming or transferring a covered repository is safe.** Grants track the repository itself, not its name.
+- **Fair use is a monthly ceiling**, shared across the repositories in your grant. Going over it doesn't block anything — reviews simply fall back to your free tier and balance for the rest of the month.
+- **Grants are time-limited** and renewed on request. When one expires, reviews fall back to the standard free tier and balance — they are never cut off outright.
+
+If a grant lapses or a project outgrows the fair-use limit, MergeWatch will say so on the pull request and point you at renewal or at [self-hosting](/self-hosting/overview), which is free and unlimited. You will not be asked for a card.
+
 Self-hosting is always free regardless, and remains the right choice if you would rather run the whole pipeline yourself. See [Self-hosted is always free](#self-hosted-is-always-free).
 
 ## Per-review pricing
--- a/docs/feat/20260812-01-oss-program-entitlement.plan.md
+++ b/docs/feat/20260812-01-oss-program-entitlement.plan.md
@@ -98,9 +98,9 @@ Ships dark — nothing writes the fields yet, so this is a no-op in prod and saf
 
 ### Phase 2 — Webhook plumbing + gate wiring
 
-- [ ] Forward `repoId` + `isPublic` from `webhook.ts:230` through the review-agent event.
-- [ ] Pass `repoContext` at `review-agent.ts:450`.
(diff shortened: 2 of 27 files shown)


santthosh:
## Review triage

Verified all three findings against the code. **One applied, two rejected with evidence.**

### ✅ Applied — `ossReason` self-describing (`billing-check.ts:38`)

Fixed in b72c270. The value was never actually ambiguous — `undefined` could only mean "OSS was not evaluated", because a supplied repo context always yields a concrete reason from `evaluateOssGrant`. But the finding is right that a reader shouldn't have to derive that. It now carries the concrete `'no_repo_context'` the predicate already returns, which also drops a ternary. `isLapsedOssGrant` already returns false for it (tested), so block copy is unchanged.

### ❌ Rejected — "OSS repo context constructed from untrusted webhook data" (`review-agent.ts:453`)

The premise is incorrect: this data is not untrusted.

1. **The webhook signature is verified.** `webhook.ts:515` rejects any payload failing HMAC verification with a 401 before any job is enqueued. `isPublic` derives from `repository.private` in that GitHub-signed payload — it's GitHub's own assertion about the repo, not caller input.
2. **review-agent is not externally reachable.** It's invoked by the webhook Lambda via the Lambda Invoke API (`webhook.ts:127`, `InvocationType.Event`). The `ReviewJobPayload` is constructed server-side in our own handler.

So the stated attack requires either forging a GitHub webhook signature, or holding IAM permission to invoke the review-agent Lambda directly — and anyone with the latter can do considerably worse than obtain free code reviews.

The suggestion to "add logging when repo context is used for billing decisions" is already satisfied: `billing-check.ts` logs `[billing] allow install=… reason=oss repo=…` on the sponsored path and `[billing] oss not applied … reason=…` otherwise.

Worth noting the design *does* take the underlying concern seriously in the place that matters — visibility is re-read from the live payload on **every** review rather than snapshotted at grant time, precisely so a repo flipped private can't keep drawing sponsorship.

### ❌ Rejected — "Billing recording failure silently proceeds" (`review-agent.ts:1036`)

Flagged unverified by MergeWatch's own verifier, and it doesn't hold:

1. **It is not silent.** `review-agent.ts:1046` emits `console.error('[billing] REVENUE LEAK: …')` with installation, repo, PR and cost — deliberately at ERROR so CloudWatch alarms can catch it.
2. **The tradeoff is deliberate and documented** in the comment directly above: the review comment is already posted, so throwing would re-run the entire review pipeline, which is worse than a missed billing record.
3. **Out of scope.** This is pre-existing code untouched by this PR.

The related *real* risk in this area is the unpriced-model gap I filed separately as #264 — where `estimatedCostUsd` is null, `recordReview` is skipped entirely, and that path has no error log at all.

PR #267, opened 2026-08-13
feat(dashboard): OSS Program state on the billing page + docs (#261, stage 4)

santthosh:
**Stage 4 of 4** — the capstone for #261. **Stacks on #266 → #265 → #263.**

## What this ships

- `/billing/status` returns an `oss` object — `null` for any installation without a grant.
- The billing page shows covered repositories, sponsored-this-month against the fair-use bar, and sponsored-to-date. An expired grant renders explicitly with a renewal link, not a payment prompt.
- `docs-site/saas/billing.mdx` describes how a grant actually behaves.
- `docs/pending/oss-program.md` graduates to `docs/oss-program.md`; E2E-82/83 marked shipped.

## Two deliberate departures from the plan

**1. The OSS panel sits *alongside* the credit-balance card, not replacing it.** The plan said "replacing the balance/top-up prompt", and that's wrong for the data model: a granted installation can still hold private or unnamed repos that bill normally, so both states are simultaneously true. Showing only the OSS panel would imply the whole installation is free. The panel says so explicitly — *"Only these repositories are sponsored, and only while they are public. Anything else in this installation bills normally."*

**2. `active` is computed server-side.** It would have been easy to ship `expiresAt` and let the client compare dates, but then the dashboard has its own second opinion about what "active" means and can drift from `evaluateOssGrant`. Same reason `sponsoredThisPeriodCents` reports 0 once `ossPeriod` goes stale — it mirrors how the cap is actually applied rather than showing a figure the gate ignores.

## Docs correction

The previous `billing.mdx` text implied a blanket free tier for OSS projects. The new section states the real constraints: named repos only, public-at-review-time, rename-safe, fair use as a soft ceiling, and expiry falling back to the free tier rather than cutting off. It also promises the thing the code now guarantees — a lapsed grant never asks a maintainer for a card.

## Verification

Build ✅ typecheck ✅ full suite ✅. Dashboard production build passes.

## Merge order

#263 → #265 → #266 → this. Each is independently reviewable; the stack must merge bottom-up.

### DIFF
--- a/docs-site/saas/billing.mdx
+++ b/docs-site/saas/billing.mdx
@@ -21,6 +21,20 @@ MergeWatch offers **free frontier-model review to open-source maintainers**, bey
 Applications are reviewed individually — tell us about the project and how you'd use MergeWatch.
 </Card>
 
+### How an approved grant behaves
+
+Once a project is approved, reviews on the covered repositories are **sponsored**: no charge, and they don't consume your free reviews or credit balance. Your **Dashboard → Billing** page shows an Open Source Program panel with the covered repositories, what's been sponsored this month, and your fair-use headroom.
+
+Specifics worth knowing:
+
+- **Only the repositories named in the grant are covered.** Approval is per project, not per organization — other repositories in the same installation bill normally. Ask us to add a repository and we will.
+- **Coverage requires the repository to be public at review time.** If a covered repository is made private, its next review is billed normally rather than sponsored.
+- **Renaming or transferring a covered repository is safe.** Grants track the repository itself, not its name.
+- **Fair use is a monthly ceiling**, shared across the repositories in your grant. Going over it doesn't block anything — reviews simply fall back to your free tier and balance for the rest of the month.
+- **Grants are time-limited** and renewed on request. When one expires, reviews fall back to the standard free tier and balance — they are never cut off outright.
+
+If a grant lapses or a project outgrows the fair-use limit, MergeWatch will say so on the pull request and point you at renewal or at [self-hosting](/self-hosting/overview), which is free and unlimited. You will not be asked for a card.
+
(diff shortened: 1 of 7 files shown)


santthosh:
## Review triage

Both findings verified against the code. **One applied, one rejected.**

### ✅ Applied — division by zero in the fair-use bar (`BillingClient.tsx:282`)

Real bug, fixed in a497059. Reproduced first:

```
cap=0, spent=0 -> width: NaN%     ← invalid CSS; the bar silently disappears
cap=0, spent=5 -> width: 100%     ← fine
```

`0/0` is `NaN`, and `Math.min(100, NaN)` is `NaN`, so the guard I had didn't catch it. Reachable in practice: `grant-oss.ts --cap 0` passes `Number("0")` straight through, and the panel's guard was `!== null`, which `0` satisfies.

Fixed at two layers:
- The bar treats a zero cap as fully consumed rather than dividing by it.
- **`grant-oss.ts` now rejects `--cap <= 0` and `--months <= 0`** — the root fix. A zero cap produces a grant that is *active but sponsors nothing*, which is the most confusing state a maintainer could be shown; `--revoke` is how you end a grant. That validation logically belongs to stage 3 (#266) but lands here rather than restacking four branches for a two-line guard.

### ❌ Rejected — `Date.parse()` NaN handling (`billing.ts:242`)

The description is accurate but the criticism inverts the intent. The code already guards:

```ts
active: Number.isFinite(expiresAt) && expiresAt > Date.now(),
```

A malformed expiry yields `active: false` — **fail-closed, deliberately**, and identical to what the billing gate does (`evaluateOssGrant` also treats an unparseable expiry as expired, with a test pinning that). Keeping the dashboard and the gate in agreement is the specific thing this function was written for; the PR body calls it out.

The suggested `return null` would be worse: a grant with a corrupt expiry would render **no panel at all**, leaving the operator with no signal that anything is wrong. Showing the panel with an "Expired" badge surfaces the anomaly instead of hiding it.

Build, typecheck, and the full suite are green after the fix.

Issue #262, opened 2026-08-13
Default model: migrate Opus 4.6 → Claude Sonnet 5 (cost) — requires sampling-param + pricing-row changes first


santthosh:
## Goal

Move the default review model from **Claude Opus 4.6** to **Claude Sonnet 5**, to cut per-review LLM cost.

Current defaults:

| Where | Value |
|---|---|
| SaaS Lambda | `us.anthropic.claude-opus-4-6-v1` (`packages/lambda/src/handlers/review-agent.ts:103`) |
| Core config | `us.anthropic.claude-opus-4-6-v1` (`packages/core/src/config/defaults.ts:186`) |
| Light model | `us.anthropic.claude-haiku-4-5-20251001-v1:0` (`defaults.ts:187`) — unchanged by this issue |
| Self-hosted | `LLM_MODEL` unset in `.env.example`; falls through to the core default |

## Why this is worth doing beyond cost

Opus 4.6 is three releases behind the current Opus line (4.7 → 4.8 → Opus 5). Claude Sonnet 5 substantially improves on Sonnet 4.6 for coding and agentic work and reaches what was previously Opus-tier quality on many tasks — so against a three-generation-old Opus this is plausibly **neutral-to-better on review quality**, not the quality-for-cost trade the tier names suggest. That should still be measured (see Acceptance criteria), not assumed.

## Cost: real, but smaller than the sticker delta

| | Input $/1M | Output $/1M |
|---|---|---|
| Opus 4.6 (today) | $5 | $25 |
| Sonnet 5 | $3 ($2 intro through 2026-08-31) | $15 ($10 intro) |

That is a 40% per-token cut (60% at intro pricing) — **but Sonnet 5 uses the newer tokenizer** (the one introduced with Opus 4.7), which Opus 4.6 does not. The same text tokenizes to roughly 1×–1.35× as many tokens. Net saving is therefore somewhere between ~19% and ~40% depending on our actual diff/prompt shape, not the headline 40%.

**Re-baseline with `count_tokens` on representative review prompts before committing to a number.** Note Bedrock pricing is set by AWS and differs from the first-party rates above — confirm against the Bedrock pricing page for the final figures.

---

### 1. Sampling parameters return a 400

`packages/llm-bedrock/src/bedrock-provider.ts:47` sends `temperature: sampling.temperature ?? 0` on **every** request, plus optional `top_p` (line 50) and `top_k` (line 51).

[...]

PR #269, opened 2026-08-13
feat(model): default to Opus 4.6 everywhere; unblock the Sonnet 5 migration (#262)

santthosh:
Toward #262. **Stacks on #268** (#264), which made the deployed model a one-line edit in `infra/params/*.env`.

This is everything #262 needs before the model can change — **minus the change itself**, which is blocked on an AWS account action (below).

---

## Sampling parameters are now per-model, not removed

Anthropic removed `temperature` / `top_p` / `top_k` from Opus 4.7 onward and the entire 5 generation; sending them returns a **400**, not a degraded response. `bedrock-provider.ts` sent `temperature: sampling.temperature ?? 0` on *every* request, so flipping the model would have failed every review.

**Deleting them outright was the wrong fix.** Prod runs Sonnet 4, which accepts them, and MergeWatch has always sent `temperature: 0` for deterministic reviews. A blanket removal would have silently moved production reviews from deterministic to default sampling — a behavior change with no error to notice.

So the provider asks whether the target model accepts them. The patterns are deliberately narrow, because the near misses are the trap:

| ID | Accepts sampling? |
|---|---|
| `claude-sonnet-5`, `claude-opus-5`, `claude-opus-4-7/4-8` | ❌ rejects |
| `claude-haiku-4-5` | ✅ — contains a "5", not the 5 generation |
| `claude-sonnet-4-5` | ✅ — same trap |
| `claude-sonnet-4-20250514` (prod today) | ✅ — unchanged |
| `claude-opus-4-6` (dev today) | ✅ — unchanged |

Unknown models default to **accepting**, so adopting a future rejecting model fails loudly with a 400 rather than silently dropping determinism. 20 test cases cover both directions and every near miss.

## Sonnet 5 / Opus 5 pricing — and the leak behind it

Rows added for both, in Bedrock (`us.` and `global.` profiles) and direct-Anthropic forms.

This isn't cosmetic. `estimateCost` returns `null` for an unpriced model, and `review-agent` gates `recordReview` on `estimatedCostUsd != null`. **Changing the default to a model with no pricing row would have skipped billing entirely** — no free-tier increment, no balance deduction, no error, every review free.

The existing `REVENUE LEAK` log only fires when `recordReview` *throws*, never when it's skipped. Added an explicit `UNPRICED MODEL` error for the skip path so the condition is alarmable instead of invisible.

---

## Not done: the flip itself — blocked on a console action

Sonnet 5 **cannot be invoked by this account yet**. Verified directly against Bedrock:

```
us.anthropic.claude-sonnet-5        ACTIVE inference profile
invoke-model                        "anthropic.claude-sonnet-5 is not available for this account"
get-foundation-model-availability   agreementAvailability: NOT_AVAILABLE
                                    authorizationStatus:   AUTHORIZED
                                    entitlementAvailability: AVAILABLE
                                    regionAvailability:    AVAILABLE
```

A control invoke against the current prod model succeeded *with* `temperature`, so the request shape is correct and IAM is fine. The sole blocker is the un-accepted **model agreement** — Bedrock console → *Model access* → enable Anthropic Claude Sonnet 5. Viewing the model-catalog page does not accept it.

Once accepted, the migration is one line in `infra/params/prod.env`, and #268's post-deploy verification step confirms it took.

## Still to decide before flipping (documented in #262)

- **Adaptive thinking is on by default on Sonnet 5**, and `maxTokensPerAgent` is 4096 — thinking and response share that budget, so agent JSON can truncate mid-object. Needs either `thinking: {type: "disabled"}` or a raised cap, A/B'd on real PRs.
- **The new tokenizer** produces ~30% more tokens for the same text, so cost should be re-baselined rather than assumed from the sticker price.

## Verification

Build ✅ typecheck ✅ full suite ✅ — 34 llm-bedrock (20 new), 827 core, 64 lambda, 123 billing, 95 server.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -198,7 +198,18 @@ jobs:
             echo "Smoke invoke succeeded — $ACTUAL is reachable and accepts the request shape."
           else
             echo "::error::Deployed model $ACTUAL is configured but NOT invocable. Reviews would fail on every PR."
-            sed -n '1,20p' /tmp/smoke-err.txt
+            # Print the whole error. A Bedrock refusal is a few lines, and
+            # silently clipping it hides the one thing worth reading — an
+            # AccessDeniedException reads very differently from a
+            # ValidationException. Guarded with an if (not `[ ] && echo`, whose
+            # non-zero exit would itself fail the step under bash -e).
+            ERR_LINES=$(wc -l < /tmp/smoke-err.txt)
+            if [ "$ERR_LINES" -gt 60 ]; then
+              sed -n '1,60p' /tmp/smoke-err.txt
+              echo "... ($((ERR_LINES - 60)) more lines suppressed)"
+            else
+              cat /tmp/smoke-err.txt
+            fi
             exit 1
           fi
 
@@ -328,7 +339,18 @@ jobs:
             echo "Smoke invoke succeeded — $ACTUAL is reachable and accepts the request shape."
           else
             echo "::error::Deployed model $ACTUAL is configured but NOT invocable. Reviews would fail on every PR."
-            sed -n '1,20p' /tmp/smoke-err.txt
+            # Print the whole error. A Bedrock refusal is a few lines, and
+            # silently clipping it hides the one thing worth reading — an
+            # AccessDeniedException reads very differently from a
+            # ValidationException. Guarded with an if (not `[ ] && echo`, whose
+            # non-zero exit would itself fail the step under bash -e).
+            ERR_LINES=$(wc -l < /tmp/smoke-err.txt)
+            if [ "$ERR_LINES" -gt 60 ]; then
+              sed -n '1,60p' /tmp/smoke-err.txt
+              echo "... ($((ERR_LINES - 60)) more lines suppressed)"
+            else
+              cat /tmp/smoke-err.txt
+            fi
             exit 1
           fi
 
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -179,6 +179,29 @@ jobs:
           fi
           echo "Review model matches the repo."
 
+          # Matching the parameter is not the same as the model being usable.
+          # A Bedrock model can be listed ACTIVE, have its agreement accepted,
+          # and still reject InvokeModel for minutes afterwards — or forever, if
+          # the account never accepted the agreement at all. Without this probe
+          # the deploy goes green and every review 400s instead.
+          #
+          # The body deliberately mirrors what the provider sends: no sampling
+          # parameters, because Opus 4.7+ and the 5 generation reject them
+          # (see acceptsSamplingParams in packages/llm-bedrock).
+          echo '{"anthropic_version":"bedrock-2023-05-31","max_tokens":8,"messages":[{"role":"user","content":"ping"}]}' > /tmp/smoke.json
+          if aws bedrock-runtime invoke-model \
+               --model-id "$ACTUAL" \
+               --region "$AWS_REGION" \
+               --body fileb:///tmp/smoke.json \
+               --cli-binary-format raw-in-base64-out \
+               /tmp/smoke-out.json > /dev/null 2>/tmp/smoke-err.txt; then
+            echo "Smoke invoke succeeded — $ACTUAL is reachable and accepts the request shape."
+          else
+            echo "::error::Deployed model $ACTUAL is configured but NOT invocable. Reviews would fail on every PR."
+            sed -n '1,20p' /tmp/smoke-err.txt
+            exit 1
+          fi
+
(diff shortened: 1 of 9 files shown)


santthosh:
## Scope update: this now also moves prod to Opus 4.6

Originally this PR was prep-only, deliberately leaving prod on Sonnet 4. That has changed at your request — **prod now defaults to Opus 4.6, matching dev.**

### What that fixes

Prod sat on Sonnet 4 only because the prod deploy step omitted `DefaultBedrockModelId` while dev passed it (#264). Dev has therefore never exercised what prod runs. Both stages now declare the same model, which is also what `template.yaml`'s parameter Default, `DEFAULT_CONFIG.model`, and this repo's own `.mergewatch.yml` have always claimed.

### Verified, not assumed

The Sonnet 5 episode was a lesson in not trusting availability signals, so Opus 4.6 was checked with a live `InvokeModel` — it succeeds, and succeeds **with `temperature`**, which empirically confirms this model's classification in `acceptsSamplingParams` rather than relying on documentation. That's the first real-API validation any part of the capability check has had.

### Cost

**~67% more per token** (/ → / per 1M). A deliberate quality-for-cost trade, not a side effect. Sonnet 5 is both more capable *and* cheaper (.20/) and remains the destination — it just isn't reachable on this account.

### One thing I could not verify

Opus 4.6's **true** Bedrock rate. Its agreement is already accepted, so `list-foundation-model-agreement-offers` returns no rate card for it. Our pricing table carries $5/$25 from Anthropic's list prices — but both cards I *could* read showed the `us.` cross-region profile priced ~10% above list:

| Model | global | us. |
|---|---|---|
| Sonnet 5 | $2.00 / $10.00 | $2.20 / $11.00 |
| Opus 5 | $5.00 / $25.00 | $5.50 / $27.50 |

If that pattern holds, `us.anthropic.claude-opus-4-6-v1` actually costs $5.50/$27.50 and we **under-bill by 10%**. This is pre-existing — dev has run this model against this pricing row all along — but promoting it to the prod default widens the blast radius. Worth confirming against Cost Explorer before or shortly after merge.

### Sonnet 5 status

Still `AccessDeniedException` with every availability field reporting `AVAILABLE`, across many hours. Blocker documented inline in `infra/params/dev.env`; flipping either stage stays a one-line change.

santthosh:
## Review triage

Four findings. **Three applied, one rejected** — with a test pinning why, rather than just disagreeing. Pushed as c6be3d1.

### ❌ Rejected — "fallback case may pass wrong modelId to buildAnthropicBody" (`bedrock-provider.ts:130`)

The concern is that the non-Anthropic fallback path now runs `acceptsSamplingParams()` on a non-Anthropic id. It does — but every reject pattern requires a `claude-` prefix, so any such id resolves to **accepts**, which is byte-for-byte the pre-#262 behavior of always sending sampling params. Verified:

```
meta.llama3-70b-instruct-v1:0     sampling included: true   (pre-#262: true)
cohere.command-r-plus-v1:0        sampling included: true   (pre-#262: true)
mistral.mistral-large-2402-v1:0   sampling included: true   (pre-#262: true)
some-unregistered-model           sampling included: true   (pre-#262: true)
```

Zero behavior change. That said, the finding points at a real latent risk — if those patterns were ever broadened, the fallback path *could* start stripping sampling from a third-party model. So the invariant is now covered by tests rather than left as an argument.

### ✅ Applied — smoke-test error output clipped at 20 lines (`deploy.yml:201`, `:328`)

Correct, and more consequential than it looks. This branch spent hours on a Bedrock refusal where `AccessDeniedException` vs `ValidationException` was the entire diagnosis — silently losing the tail of that error hides the one thing worth reading. Now prints the full error, with an explicit count when it exceeds 60 lines.

**One note on the suggested fix**, which I did not use:

```bash
sed -n '1,20p' file && [ $(wc -l < file) -gt 20 ] && echo '... (truncated)'
```

As the final command in a `run` block this fails the step whenever the file is ≤20 lines — `[ ]` exits non-zero, and GitHub Actions runs with `bash -e`. So the suggestion would turn every *successful* smoke test into a failed deploy. Used an `if` instead.

### ✅ Applied — comment lists models the regex doesn't match (info, `bedrock-provider.ts:52`)

Accurate: the regex alternation includes `haiku` while the comment above listed only sonnet/opus/fable/mythos. The comment now explains that `haiku` is present for symmetry with a hypothetical Haiku 5, and that the negative lookahead is what keeps `claude-haiku-4-5` out of the pattern.

---

Build, typecheck, and the full suite are green — 39 llm-bedrock tests (5 new).

**Note this review predates the pricing work on this branch.** It scanned 362 lines at the earlier commit and did not see the Bedrock `us.` cross-region correction (de4d26b), which is the highest-consequence change here — it raises billed cost on every review. Worth a re-review before merge.

Issue #253, opened 2026-08-11
Use repo-attached Agent Skills as review context


santthosh:
**Source:** user suggestion — MergeWatch should take advantage of the skills checked into a repo, since many of them already encode that team's best practices.

## Why this fits

MergeWatch already has the mechanism; it's just pointed at one file. `fetchConventions()` in `packages/core/src/config/conventions.ts` resolves exactly one source, first hit wins:

1. `conventions:` in `.mergewatch.yml`
2. `AGENTS.md`
3. `CONVENTIONS.md`
4. `.mergewatch/conventions.md`

Later candidates are never fetched, and content is capped at 16 KB.

That captures repo-wide rules, but teams increasingly put *task-specific* best practices in skills instead — how to write a migration, how to add an endpoint, what a good test looks like here. Those are exactly the standards a reviewer should hold a PR to, and MergeWatch currently can't see any of them.

We already dogfood this. Both forms are checked into this repo:

```
.claude/skills/resolve-mergewatch.md          # flat file
.claude/skills/ship-feature/SKILL.md          # directory form
```

`ship-feature/SKILL.md` encodes real house process — phased plan docs, stacked PRs, e2e RUNBOOK scenarios, gating each stage on build+typecheck+test. A reviewer that knew that could hold PRs to it. Today it reviews blind to it.

## Proposal

Discover repo-attached skills and make them available as review context, alongside conventions rather than replacing them.

The `SKILL.md` format is already built for this. Frontmatter carries a `description` written specifically for relevance matching:

```yaml
---
description: Drive a feature or non-trivial change from request to shipped — write a phased plan doc in docs/feat or docs/chore, get it approved, then implement in stacked PRs...
---
```

That maps cleanly onto MergeWatch's existing codebase-awareness path: inject the cheap `name` + `description` index for every discovered skill, and let an agent request the full body only when a skill looks relevant to the diff. Same progressive-disclosure shape skills were designed around, and it keeps the common case to roughly one extra API call.

## Design questions

**Discovery.** `.claude/skills/` is the obvious start. Worth deciding whether to also scan `.agents/skills/`, a bare `skills/`, or make it configurable via a `skills:` key in `.mergewatch.yml` mirroring `conventions:`. Both the flat `<name>.md` and directory `<name>/SKILL.md` layouts need handling — this repo has one of each.

[...]

PR #306, opened 2026-08-14
Make the committed .mergewatch.yml win over dashboard settings

brunomgurgel:
## What

The configuration docs promise that a repo's committed `.mergewatch.yml` always wins over dashboard settings and built-in defaults ("the .mergewatch.yml in the repository always wins"). Both review paths currently do the opposite: `settingsOverrides` (built from the dashboard InstallationSettings) is spread **after** `yamlConfig`, so dashboard values silently clobber `maxFindings`, the per-agent toggles, `customStyleRules` and `minSeverity` for any repo that sets them in the yml.

- `packages/server/src/review-processor.ts` and `packages/lambda/src/handlers/review-agent.ts`: precedence is now defaults < dashboard settings < stored installation config < committed yml < `LLM_MODEL` env pin. The `agents` object gets a one-level merge so dashboard toggles still apply for agents the yml doesn't mention.
- `packages/core/src/agents/reviewer.ts`: the agentic-fetch success log now names the fetched files (`Agent fetched 2 file(s) in 2 round(s): path/a.md, path/b.md`). We spent a while diagnosing fetch behavior from deployment logs, and the count without paths made it much slower than it needed to be.

## Tests

Two new cases in `review-processor.test.ts`: yml-set keys (`maxFindings`, `customStyleRules`) beat dashboard values; dashboard values (including per-agent toggles) apply where the yml is silent, and a yml agent toggle beats the dashboard's. `pnpm vitest run` green: server 48, lambda 64, core 818.

## One observation, deliberately not changed here

`minSeverity` is parsed from the yml and merged into the runtime config, but nothing in the runtime consumes it — findings are never filtered by severity. Worth either wiring it into the findings filter or removing it from the docs; happy to follow up with either once you pick a direction.

Found while configuring conventions-driven reviews on a large monorepo — same motivation as #253.


santthosh:
Reviewed this alongside #268/#269, which touch the same config-resolution code. **The precedence change is sound and the MergeWatch finding on it is a false positive** — no deep merge needed.

## On "shallow merge of agents configuration may lose nested settings"

The warning is only valid if `agents` contains nested objects. It doesn't — the type is eight flat booleans (`packages/core/src/config/defaults.ts:126-135`):

```ts
agents: {
  security: boolean;      bugs: boolean;
  style: boolean;         summary: boolean;
  diagram: boolean;       errorHandling: boolean;
  testCoverage: boolean;  commentAccuracy: boolean;
}
```

For a flat boolean map, a deep merge is byte-for-byte identical to the one-level spread you wrote. Adding `deepMerge` here would introduce a helper that provably does nothing.

Your one-level merge is also the *correct* level — it's exactly what preserves a dashboard toggle for agents the yml is silent about:

```
dashboard: { security: false, bugs: true, style: true, testCoverage: true, ... }
yml:       { style: false, testCoverage: false }
result:    { security: false, bugs: true, style: false, testCoverage: false, ... }
            ^^^^^^^^^^^^^^^^ dashboard survives    ^^^^^^^^^^^^^^^^ yml wins
```

So: no change needed, and I'd push back on the suggestion rather than spend time on it.

## Heads-up: your `review-processor.ts` diff is unreviewable, and it's not your fault

That file renders as `Bin 44577 -> 45080 bytes` rather than a diff. Cause: it contains **two raw NUL bytes** at line 50, which make git classify it as binary:

```ts
const key = `${model}\x00${input ?? ''}\x00${output ?? ''}`;
```

They're a cache-key delimiter and they're **already on `main`** — pre-existing, nothing to do with this PR. But the consequence lands here: nobody can review your changes to that file through a normal diff, which likely explains why MergeWatch only flagged the `review-agent.ts` half of a change you made identically in both runtimes.

I read it by stripping the NULs, and the server-side change matches the Lambda one correctly. Worth fixing separately (`\0` escapes instead of literal bytes) so a 45KB file on the self-hosted review path becomes reviewable again.

## One cross-PR interaction to be aware of

#268 (open, mine) changes model resolution to `installation.modelId ?? yamlConfig.model ?? deployDefault` — which has the **stored installation config beating the committed yml**, the reverse of the rule you're establishing here. No behavioral difference today (nothing writes `installation.modelId`; all 525 installation rows have it unset), so it's latent rather than broken. But once both land, two adjacent paths in `review-agent.ts` will disagree about which source wins.

Yours is the documented contract, so I'll align mine to it. Flagging so it's visible rather than discovered later.

## On `minSeverity`

Good catch, and it's the third instance of the same pattern we've hit this week: `config.model` was parsed but never read (#264), `installation.config` is empty on all 525 rows, and now `minSeverity` is merged but never filters anything. Might be worth one issue covering documented-but-inert config keys rather than three separate discoveries.

PR #307, opened 2026-08-16
fix(server): replace literal NUL bytes so review-processor.ts is diffable

santthosh:
## The problem

`packages/server/src/review-processor.ts` contains two **literal NUL bytes** at line 50, used as a cache-key delimiter.

A NUL byte makes git classify a file as binary. Every diff to this 45KB file — which sits on the entire self-hosted review path — has rendered as:

```
packages/server/src/review-processor.ts | Bin 44577 -> 45080 bytes
1 file changed, 0 insertions(+), 0 deletions(-)
```

Not a patch. Unreviewable on GitHub, in `git diff`, and **to MergeWatch's own reviewer, which reads diffs.**

## It already cost us a review

This surfaced on #306, an external contributor's PR touching this file. The diff was an undiffable blob, and the MergeWatch review that ran against it flagged only the half of the change living in a *different* file — the identical server-side change was invisible to it. I had to strip the NULs locally just to read what the PR did.

So the practical cost: **any change to this file has been shipping without meaningful diff review**, by anyone, for as long as the bytes have been there.

## The fix

Escape sequences instead of literal bytes. Byte-for-byte identical at runtime — verified, the key still resolves to char codes `109,0,105,0,111` — and the file stays text.

Confirmed the outcome rather than assuming it. After the fix, a scratch edit diffs as:

```
packages/server/src/review-processor.ts | 2 ++
+
+// scratch
```

## The guard

Adds a repo-wide `source-hygiene.test.ts` scanning all text sources for NUL bytes.

**Written first and confirmed to fail on the unfixed tree**, naming the file, byte count and line:

```
× no text source file contains a NUL byte
+ "packages/server/src/review-processor.ts — 2 NUL byte(s), first at line 50"
```

A guard that has only ever been seen passing isn't known to catch anything. This one is.

The scan covers the whole monorepo (skipping `node_modules`, `dist`, `.next`, `.turbo`, etc.) and found exactly one offender, so no wider cleanup is hiding behind this.

## Scope

Deliberately narrow: the byte fix plus the guard. No behavior change, no refactor of the surrounding pricing-warning logic.

Independent of #269 and #306 — touches a line neither modifies, so it can merge in any order.

## Verification

Build ✅ typecheck ✅ full suite ✅.


santthosh:
## Why GitHub still showed this file as binary

Worth writing down, because it is counterintuitive: **it was not stale, and not this branch's fault.**

Git decides a blob is binary if it finds a NUL byte within the first 8000 bytes, and it marks a *diff* binary when **either side** is. Measured on this PR:

| Side | NUL bytes |
|---|---|
| base (`origin/main`) | **2**, first at byte 2596 |
| head (this branch) | 0 |

So the commit that removes the NULs cannot render its own diff as text — the "before" side is binary by definition. `numstat` shows it exactly:

```
main..PR       -   -    (binary)
PR..PR+edit    2   0    (text)
```

Every diff *after* this merges would have been text. But that means the one change most worth reading — the fix itself — was the one nobody could read.

## Fixed properly

Added `.gitattributes` setting the `diff` attribute on source types, which forces a textual diff regardless of content. The PR now renders as:

```
33  0  .gitattributes
 1  1  packages/server/src/review-processor.ts
```

One line changed, visible and reviewable.

This **complements** the guard test rather than replacing it:

- `source-hygiene.test.ts` stops a NUL byte being committed at all.
- `.gitattributes` keeps the file readable if one ever is.

Belt and braces is justified here specifically because the failure mode hides itself — an undiffable file looks fine until someone tries to review it, and by then the review has already happened against a blob.

Build, typecheck, and the full suite remain green.

PR #313, opened 2026-08-16
fix(assets): make the wordmark legible in dark mode

santthosh:
## The bug

The wordmark set `color="#0D1117"` on the root `<svg>` and drove every element through `currentColor`:

```svg
<svg … color="#0D1117">
  .wm { fill: currentColor; }
  <path  stroke="currentColor" />
  <circle fill="currentColor" />
  <circle fill="#F4F2EA" />   ← opaque cream highlight
```

So the mark and text rendered near-black regardless of the surrounding theme, and on any dark background they disappeared.

That includes **MergeWatch's own PR comments** — `comment-formatter.ts:268` embeds this file from `raw.githubusercontent`, so every dark-mode user has been seeing an empty gap at the top of every review comment.

The lens highlight compounded it: a solid `#F4F2EA` circle painted over the pupil reads as a cream dot floating on dark, not as a highlight.

## The fix

- **Highlight is now a mask**, not an opaque circle — it punches a transparent hole, so it picks up whatever is behind it on any background.
- **Colors are explicit and theme-neutral** — `#16A34A` for the mark and the `.ai` suffix, `#767C87` for the wordmark text — rather than inheriting a hardcoded near-black.

## Two things that came along with it, worth calling out

**The Google Fonts `@import` is gone**, replaced by an inline `font-family` with system fallbacks. GitHub sanitises embedded SVGs and never loaded that font, so it was a network request that bought nothing — and it leaked a third-party request for anyone viewing the README.

**`viewBox` narrows 1000 → 940.** I checked both consumers: `README.md` and the PR-comment footer both set `height="48"` and no width, so it scales proportionally and nothing reflows.

## Not changed

`packages/dashboard/components/MergeWatchLogo.tsx` is a separate React port of the same geometry and already adapts — `currentColor` for the mark, `var(--bg-page)` for the highlight. It needs nothing here.

Worth noting for later: the two will now differ in intent — the SVG is explicitly green-marked, while the React component inherits the surrounding text color. That's arguably correct for each context, but if brand consistency matters, the component is where to reconcile it. Its doc comment also still claims the highlight is "fixed at #F4F2EA per the brand spec" while the code uses `var(--bg-page)`.

## Verification

Parses as valid XML; `viewBox 0 0 940 180`; `<title>mergewatch.ai</title>` added for assistive tech alongside the existing `role`/`aria-label`; no external references remain. Full suite green (834 core), including the `comment-formatter` test that asserts the wordmark URL.

### DIFF
--- a/assets/mergewatch-wordmark.svg
+++ b/assets/mergewatch-wordmark.svg
@@ -1,15 +1,14 @@
-<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 1000 180" width="1000" height="180" role="img" aria-label="Mergewatch" color="#0D1117">
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 940 180" width="940" height="180" role="img" aria-label="mergewatch.ai">
+  <title>mergewatch.ai</title>
   <defs>
-    <style type="text/css">
-      @import url('https://fonts.googleapis.com/css2?family=Nunito:wght@800&amp;display=swap');
-      .wm { font-family: 'Nunito', system-ui, -apple-system, sans-serif; font-weight: 800; letter-spacing: -0.02em; fill: currentColor; }
-      .wm-dim { opacity: 0.55; }
-    </style>
+    <mask id="mw-pupil">
+      <circle cx="512" cy="644.4" r="108" fill="#fff"></circle>
+      <circle cx="474.2" cy="601.2" r="32.4" fill="#000"></circle>
+    </mask>
   </defs>
-  <g transform="translate(120 21.599999999999994) scale(0.13359375)">
-    <path d="M 207 579.4 Q 512 274.4 817 579.4" fill="none" stroke="currentColor" stroke-width="82" stroke-linecap="round" />
-    <circle cx="512" cy="644.4" r="108" fill="currentColor" />
-    <circle cx="474.2" cy="601.1999999999999" r="32.4" fill="#F4F2EA" />
+  <g transform="translate(120 21.6) scale(0.13359375)" color="#16A34A">
+    <path d="M 207 579.4 Q 512 274.4 817 579.4" fill="none" stroke="currentColor" stroke-width="82" stroke-linecap="round"></path>
+    <circle cx="512" cy="644.4" r="108" fill="currentColor" mask="url(#mw-pupil)"></circle>
   </g>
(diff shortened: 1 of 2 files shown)


santthosh:
## Added: the React wordmark now matches the SVG (1aa2259)

The two had drifted into different brands. With the dark-mode fix in place the divergence became visible — the embedded SVG renders a green mark with a gray wordmark, while the dashboard rendered a `currentColor` mark with a dimmed `.ai` in a different typeface.

Aligned on the SVG as the brand reference:

| | assets SVG | MergeWatchLogo.tsx (before) | now |
|---|---|---|---|
| mark | `#16A34A` | `currentColor` | `#16A34A` |
| wordmark text | `#767C87` | inherited fg | `#767C87` |
| `.ai` suffix | `#16A34A` | `opacity-55` | `#16A34A` |
| font | Plus Jakarta Sans | Nunito | Plus Jakarta Sans |
| highlight | mask (transparent) | `var(--bg-page)` | evenodd hole (transparent) |

### A second latent bug this surfaced

The component's highlight was an opaque circle filled with `var(--bg-page)`. That only works where the mark sits **directly on the page background** — over a card, a modal, or any tinted surface it painted a page-colored dot rather than a hole. Same class of bug as the SVG's cream circle, just less obvious because it happened to match in the common case.

It is now a genuine hole.

### One deliberate difference from the SVG

The SVG punches its hole with a `mask`; the component uses `fill-rule="evenodd"`. Mask ids are document-global, and this component can render several times on one page — duplicate ids are invalid even where browsers tolerate them. The evenodd path avoids ids entirely.

Verified the two subpaths reproduce the original circles exactly:

```
circle cx=512   cy=644.4 r=108   ->  M 404,644.4   a 108,108 0 1,0 216,0   a 108,108 0 1,0 -216,0
circle cx=474.2 cy=601.2 r=32.4  ->  M 441.8,601.2 a 32.4,32.4 0 1,0 64.8,0 a 32.4,32.4 0 1,0 -64.8,0
```

### Blast radius checked, not assumed

- **No consumer passes a `text-*` class** to these components, so nothing depended on the inherited color.
- **The font swap is contained** — `nunito.className` was referenced only by the wordmark, nowhere else in the app.
- The stale doc comment claiming the highlight is *"fixed at #F4F2EA per the brand spec"* is gone; it had been describing behavior the code stopped doing some time ago.

Dashboard production build, repo build, typecheck and full suite all green.

Issue #337, opened 2026-08-16
Date-only end_date excludes the entire final day of the range


santthosh:
`packages/dashboard/app/api/analytics/route.ts:34-38`

```ts
const isoDateRegex = /^\d{4}-\d{2}-\d{2}(T\d{2}:\d{2}:\d{2}(\.\d{3})?Z)?$/;
```

The regex accepts a **date-only** value, and the stored `created_at` is `text` holding a full ISO timestamp (`packages/storage-postgres/src/schema.ts:30`). The filter is a string comparison (`dashboard-store.ts:183`):

```ts
conditions.push(lte(reviews.createdAt, endDate));
```

With `end_date=2026-08-16`, the comparison is `'2026-08-16T09:31:00.000Z' <= '2026-08-16'`, which is **false**. Every review on the final day of the range is excluded. The same applies to the Dynamo `createdAt <= :endDate` filter (`storage-dynamo/src/dashboard-store.ts:179`).

## Severity qualifier

The current dashboard UI is not affected: `AnalyticsClient.tsx:141,145-146` always sends a full ISO timestamp, with custom ranges explicitly widened to `T23:59:59.999Z`. So this is a **latent API bug**, not the cause of the reported symptom — filed to prevent it becoming one.

It matters because `/api/analytics` is a documented query surface: any direct API consumer, script, or future UI passing the obvious `?end_date=2026-08-16` silently loses a day and gets no warning.

## Fix

Normalize at the boundary: when a date-only value is supplied, expand `start_date` to `T00:00:00.000Z` and `end_date` to `T23:59:59.999Z` before it reaches the store. Apply identically to both backends.

## Related — UTC bucketing

While here, note that trend buckets are cut with `review.createdAt.substring(0, 10)` (route.ts:108, 138) — UTC calendar days — while the picker builds ranges from the viewer's local clock. For non-UTC users the daily buckets are offset from the days they selected, which shows up as activity attributed to the wrong day at the edges. Worth a decision (bucket in UTC and label it, or bucket in the viewer's zone) rather than leaving it implicit.

## Acceptance criteria

- [ ] Date-only `end_date` includes the whole final day; date-only `start_date` includes the whole first day
- [ ] Behaviour identical on Postgres and DynamoDB
- [ ] Timezone bucketing decision made and documented in the route
- [ ] Tests for date-only and full-timestamp forms

santthosh:
## 🚚 Ship tracker (auto-maintained)

[...]

Issue #336, opened 2026-08-16
p95 review duration returns the maximum for n <= 20


santthosh:
`packages/dashboard/app/api/analytics/route.ts:176-178`

```ts
const p95Duration = durations.length > 0
  ? durations[Math.floor(durations.length * 0.95)]
  : 0;
```

For a sorted ascending array of length `n`, the maximum valid index is `n - 1`, but `Math.floor(n * 0.95)` reaches `n - 1` for every `n ≤ 20`:

- `n = 20` → `floor(19) = 19` → index 19 is the **maximum**, not p95
- `n = 10` → `floor(9.5) = 9` → the maximum again
- `n = 5` → `floor(4.75) = 4` → the maximum again

So for any range with 20 or fewer completed reviews — a new instance, a single repo, a narrow date filter — **p95 duration is simply the slowest review**, reported as a percentile. It overstates the figure and it is most wrong exactly when a user is first evaluating the product.

For larger `n` it remains biased one position high (`floor(n × 0.95)` rather than the nearest-rank `ceil(n × 0.95) - 1`).

## Fix

Nearest-rank with a clamp:

```ts
const idx = Math.min(durations.length - 1, Math.max(0, Math.ceil(durations.length * 0.95) - 1));
const p95Duration = durations[idx];
```

Consider suppressing the percentile entirely below a small sample threshold — a "p95" over 4 data points is not a percentile regardless of the index arithmetic, and showing it invites false confidence.

## Acceptance criteria

- [ ] `p95` returns the correct nearest-rank element for n = 1, 5, 20, 21, 100
- [ ] Below a documented minimum sample size, the UI shows insufficient-data rather than a number
- [ ] Unit test covering the boundary cases above

Filed P1 rather than P0: it is a genuine correctness bug but affects one displayed statistic, and it overstates rather than fabricating a trend.

santthosh:
## 🚚 Ship tracker (auto-maintained)

**Scope:** Fix the p95 review-duration index (returns the maximum for n ≤ 20, biased one rank high above that) with nearest-rank arithmetic, and suppress the figure below a documented minimum sample size instead of presenting a non-percentile as one.

**Plan:** 1 PR. (The code moved since filing: #333 stage 1 extracted the computation from `route.ts` into `lib/analytics-aggregate.ts`, where it is already flagged for this issue.)

### Fix — nearest-rank p95 + insufficient-data gate ✅ MERGED  ·  PR: #346
[...]

PR #339, opened 2026-08-16
refactor(dashboard): extract analytics aggregation into a pure, tested function (#333, stage 1)

### LINKED ISSUE #333: Analytics is capped at 500 reviews and silently truncates the oldest data in the selected range
Reported on a self-hosted instance: **the analytics page tops out at 500 reviews and the trend charts do not reflect the real timeline.** Both symptoms have the same root cause.

## Root cause

`packages/dashboard/app/api/analytics/route.ts:70-72`

```ts
// Fetch up to 500 reviews for aggregation (with optional date filter)
const result = await store.reviews.listReviews(targetRepos, 500, undefined, undefined, startDate, endDate);
const reviews = result.items;
```

The page count is capped at 500 and **`result.nextCursor` is discarded** — the route never paginates. Every aggregate below it is then computed from that truncated set: `totalReviews`, `totalFindings`, `severityBreakdown`, `repoBreakdown`, `categoryBreakdown`, `statusCounts`, `costStats`, `durationStats`, `mergeScoreDistribution`, and both trend series.

## Why the timeline is wrong, not just the count

`packages/storage-postgres/src/dashboard-store.ts:186-192`

```ts
.orderBy(desc(reviews.createdAt))
.limit(limit + 1)
```

Newest first. So the 500 rows that survive are the **most recent** 500 in the selected range — and everything older is silently dropped. For any range containing more than 500 reviews, the charts render a truncated window while still being labelled with the full range: the early part of the period appears empty or flat, and `scoreTrend` / `findingsPerReviewTrend` simply have no points for those dates.

This is why widening the date range does not add history. It shifts nothing — you always get the same newest 500.

## Why it reads as a hard ceiling

`packages/dashboard/components/AnalyticsClient.tsx:594`

```tsx
<StatCard label="Total Reviews" value={data.totalReviews} />
```

`totalReviews` is `reviews.length` (route.ts:181) — the size of the truncated page. It is presented to the user as a **total**, with no indication it is a capped sample. `successRate` (AnalyticsClient.tsx:547) divides by the same capped number.

## Impact

Any repo or installation past 500 reviews in the selected window sees understated totals and a falsified history. It affects **both deployment modes** — the route is shared — and it gets worse the longer an instance runs, which is the opposite of how a metrics page should degrade.

## Fix direction

Aggregation must not be bounded by a display page size. Options, roughly in order of preference:

1. **Aggregate in the database.** These are all `SUM`/`COUNT`/`AVG`/`GROUP BY date` queries; pulling every row into Node to reduce it in a `for` loop is the actual design error. Postgres can return the trend series directly, and it removes the ceiling permanently.
2. If the in-memory shape is kept short-term, **loop `nextCursor` to exhaustion** with a hard safety bound, and if that bound is ever hit, say so in the response so the UI can label the numbers as partial.

Whatever is chosen: **never present a truncated aggregate as a total.** If a cap is hit, the UI must show it.

## Acceptance criteria

- [ ] Analytics over a >500-review range reports correct totals
- [ ] Trend series cover the full selected range, with the oldest dates present
- [ ] Widening the date range demonstrably adds older data points
- [ ] If any cap remains, the response carries a truncation flag and the UI renders it
- [ ] Regression test with >500 seeded reviews across a multi-month span — there is currently **no test file** for this route

## Related

Same family as #334 (rollup windows) and #335 (Dynamo pagination).

santthosh:
Stage 1 of 2 for #333. **No behavior change** — this PR exists to make stage 2's fix testable.

## Why

`/api/analytics` reduced reviews into ~15 aggregates inside the request handler, and there is **no API-route test harness anywhere in this repo** (`find app/api -name '*.test.ts'` returns nothing). So the logic was untestable where it lived, and three separate defects shipped through it: the 500-row truncation (#333), an off-by-one percentile (#336), and a date-boundary bug (#337).

Rather than invent a next-auth/Next.js mocking layer, the aggregation moves to `lib/` as a pure function — the convention `cost-insight.ts` already follows ("kept dependency-free so the logic is unit-testable without React") and the same shape as `buildInsightFromDispositions` in core.

## The extraction is provably a no-op

Only three deltas, all semantically identical:

| Before | After | Why |
|---|---|---|
| `(review as any).estimatedCostUsd` | `review.estimatedCostUsd` | The input is now structurally typed, so the cast is unnecessary |
| `const analytics = { … }` | `return { … }` | It's a function now |
| — | a `NOTE:` comment on the p95 index | Flags the known bug for #336 |

**The known-wrong p95 is preserved deliberately and pinned by a test.** Fixing it here would make "no behavior change" untrue and would mean this PR and #336 both touch the same line. The test asserts the current wrong value (`2000` for n=20, where nearest-rank p95 is `1900`), so #336 has an explicit assertion to flip rather than a silent behavior change.

## Input typing

`AggregatableReview` is a structural subset of `ReviewItem` rather than an import of it. That keeps the module dependency-free, lets fixtures state only the fields under test instead of satisfying two dozen irrelevant required ones, and gives `estimatedCostUsd` a real type — it's declared `number | string | null` because the DynamoDB path can return a string, which is why the original code wrapped it in `Number()`.

## Tests — 33 added

- **Zero state** — every average is `0`, not `NaN`; all buckets pre-seeded so the UI never null-checks
- **Hand-computed fixture** — totals, averages, severity/category buckets, score distribution, cost, repo ranking
- **Trends** — ascending sort regardless of input order, intra-day averaging, and **gaps stay gaps** (a synthesized zero would render as a score crash)
- **Status** — each known status counted; an unknown status ignored rather than creating a bucket; non-complete excluded from duration and cost
- **Missing/malformed fields** — `null` vs `undefined` vs absent; `mergeScore: 0` and `estimatedCostUsd: 0` counted as present, not missing (the `!= null` distinction); unknown severities counted in the total but not bucketed; string cost coerced
- **Numeric duration sort** — a default `.sort()` would order `[1000, 200, 30]` and corrupt every percentile
- **#333 regression** — a **1234-review fixture spanning ~154 days** asserting full counts, complete day coverage, ascending order across the whole span, and correct repo/cost attribution. This is the test stage 2's fix has to satisfy.

## Verification

`pnpm run build` ✅ · `pnpm run typecheck` ✅ · `pnpm run test` ✅ (33 new, full suite green)

## Next

Stage 2 removes the hardcoded `500`, paginates `nextCursor` to exhaustion, and surfaces a truncation flag in the UI. It will carry `Closes #333`.

Part of #333.

### DIFF
--- a/packages/dashboard/app/api/analytics/route.ts
+++ b/packages/dashboard/app/api/analytics/route.ts
@@ -2,6 +2,7 @@ import { NextRequest, NextResponse } from "next/server";
 import { getServerSession } from "next-auth";
 import { authOptions } from "@/lib/auth";
 import { getDashboardStore } from "@/lib/store";
+import { aggregateReviews } from "@/lib/analytics-aggregate";
 import {
   fetchUserInstallations,
   fetchAccessibleRepoNames,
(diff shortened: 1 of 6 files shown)


santthosh:
@mergewatch review

Issue #310, opened 2026-08-16
Config keys that are parsed, merged and documented — but never read (minSeverity, installation.config, installation.modelId)


santthosh:
Three config surfaces are parsed, typed, merged, and in one case documented — and then never read. Each was rediscovered independently over the past week, which is the argument for one issue rather than three.

## 1. `minSeverity` — documented, parsed, never filters anything

**Documented** in `docs-site/configuration/mergewatch-yml.mdx` as:

> Minimum severity level to report. One of `info`, `warning`, `critical`. Findings below this threshold are suppressed.

**Parsed** with validation at `packages/core/src/github/client.ts:912`. **Typed** at `packages/core/src/config/defaults.ts:162`, defaulted to `'info'` at `:211`. **Merged** into the runtime config in both runtimes, and even fed from the dashboard's `severityThreshold` (`review-agent.ts:544`, `review-processor.ts:417`).

Then nothing reads it. Grepping for a read of `minSeverity` — as opposed to an assignment — returns only the parser writing it. No finding is ever suppressed by severity.

Note the surrounding code *does* branch on `f.severity` in several places (`comment-formatter.ts:351`, `org-agents.ts:169`, `reviewer.ts:1346`), so it looks wired from a distance. It isn't: those are unrelated per-finding decisions, not a threshold filter.

Spotted by @brunomgurgel while working on #306.

- [ ] Either wire it into the findings filter, or remove it from the docs and the config surface
- [ ] If wiring: the dashboard `severityThreshold` control feeds it, so both surfaces start working at once — worth checking that's the intended behavior before shipping

## 2. `installation.config` — populated nowhere, merged everywhere

`InstallationItem.config` is typed as the parsed `.mergewatch.yml` and documented in `packages/core/src/types/db.ts` as *"fetched from the repository's default branch during installation and updated when the config file changes"*.

Measured against production: **0 of 525 installation rows have a non-empty `config` map**, including this repository's own row, which certainly has a populated `.mergewatch.yml`. Nothing writes it.

Both runtimes nevertheless merge it into the effective config — it is the `storedConfig` layer in the precedence chain #306 established (`review-processor.ts`, `review-agent.ts`). So there is a documented precedence tier that is always empty.

[...]

PR #365, opened 2026-08-17
fix(core): default severity threshold must not suppress info findings (#357)

### LINKED ISSUE #357: E2E-02 regression: info-only fixture produced zero findings — no 'Info (N)' section rendered
## Fixture
E2E-02 `02-info-only` — https://github.com/mergewatch/fixtures/pull/234

## Expected
5/5 verdict with reason "No action items — only informational notes", plus an **"Info (N)" collapsible section with ≥1 informational finding**; review Approved.

## Observed
`🟢 5/5 — Safe to merge — No issues found on changed lines.` + `🎉 All clear!`, review APPROVED — but **no "Info (N)" section exists anywhere in the comment**, and the verdict reason is "No issues found on changed lines" instead of the info-only reconciliation copy. Zero findings of any severity were emitted, so the info-only path (#134) was never exercised.

## Notes
Neither #134-downgrade nor attention-table failure modes tripped — this may be stimulus/stochastic (the LLM emitted no info notes at all) rather than a reconciliation regression, but the fixture's required outcome is concretely violated and has now missed its target. Consider strengthening the fixture bait or investigating whether info-tier findings are being dropped pre-render.

## Reproduce
`scripts/apply-fixture.sh 02-info-only`

🤖 Filed by /verify-suite

santthosh:
Closes #357.

## Root cause — not stochastic

The issue left open whether E2E-02's zero findings were stochastic ("the LLM emitted no info notes") or a real drop. It's the latter, and we introduced it: the suite ran hours after #310/#349 wired `minSeverity`, and `DEFAULT_INSTALLATION_SETTINGS.severityThreshold = 'Med'` maps to `minSeverity: 'warning'`. Inert for as long as the key was dead — the moment it went live, **every installation without saved dashboard settings silently dropped all info-tier findings pre-render**: no "Info (N)" section, "No issues found" verdicts on info-only PRs, and the #134 info-only reconciliation path never reached.

Two adjacent latents fixed in the same pass: the Postgres settings column already defaulted `'Low'` (contradicting the TS default — a row created by column default behaved differently from no-row defaults), and the server's `?? 'warning'` fallback suppressed on unknown values.

## Fix

- `DEFAULT_INSTALLATION_SETTINGS.severityThreshold`: `'Med'` → `'Low'` — out-of-the-box behavior reports every tier (the pre-#349 reality); `Med`/`High` are what they always claimed to be: explicit opt-ins. Now consistent with the Postgres column default.
- Server mapping fallback → `?? 'info'` (malformed threshold ≠ suppression). The Lambda path has no fallback and already degrades safely to the pipeline's `'info'` default.
- `SettingsForm` initial value → `'Low'`, so saving untouched settings doesn't persist an unintended suppression opt-in.

## Tests

3 regression tests pin the mapping end-to-end through the processor (default → `'info'`, explicit `Med` → `'warning'`, unknown → `'info'`) — 62/62 server suite; full workspace build/typecheck/test green (verified by exit codes). RUNBOOK E2E-78 documents the default.

## Follow-ups (out of scope here)

- Fixture-bait strengthening lives in `mergewatch/fixtures`.
- Installations that explicitly saved settings while `'Med'` was the preselected form value have a persisted opt-in they may not have intended — quantifying that in prod is blocked on an expired AWS session; noted on the tracker.

### DIFF
--- a/e2e/RUNBOOK.md
+++ b/e2e/RUNBOOK.md
@@ -2733,7 +2733,7 @@ curl -s "$MCP_URL" \
 
 ### E2E-78: Output shaping — `minSeverity`, `maxFindings`, `postSummaryOnClean`
 
-**Status:** ⬜ NOT YET COVERED. (All three knobs were documented-but-unwired at various points: `minSeverity` until #310, `maxTokensPerAgent` and `postSummaryOnClean` until #350 — all three now exist; fixture still to be run.)
+**Status:** ⬜ NOT YET COVERED. (All three knobs were documented-but-unwired at various points: `minSeverity` until #310, `maxTokensPerAgent` and `postSummaryOnClean` until #350 — all three now exist; fixture still to be run. #357: the default severity threshold is **Low** — info findings render out of the box; `Med`/`High` are explicit opt-ins, and wiring `minSeverity` with a `Med` default had silently suppressed every info-tier finding.)
 
 **Behavior:** Three independent knobs on what reaches the PR: `minSeverity` (`info` | `warning` | `critical`) drops lower-severity findings; `maxFindings` caps how many are posted; `postSummaryOnClean` decides whether a clean PR gets a comment at all.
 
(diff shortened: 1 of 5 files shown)


santthosh:
@mergewatch review

Issue #409, opened 2026-08-20
OSS Program: org-scoped grants + pre-approval for orgs that haven't installed


santthosh:
Follow-up to #261 (the OSS Program entitlement, documented in `docs/oss-program.md`).

Plan doc: `docs/feat/20260820-01-oss-org-grants-and-preapproval.plan.md` (lands with stage 1).

## Problem

The current mechanism can only express *"these specific repositories, by immutable numeric id"*. Two sharp edges in practice:

**Whole-org OSS projects have to be enumerated.** An org where everything public is open source needs each repo listed individually, and every new repo they create silently falls outside the grant until an operator notices and runs `--add`. The maintainer's experience is that MergeWatch mysteriously stopped covering half their work.

**Approval is blocked on installation.** `scripts/grant-oss.ts` resolves a repo to its installation via `GET /repos/{owner}/{repo}/installation`, so a grant cannot be written until the maintainer has installed the App. The approval email has to say "install, then tell us, then we'll grant" — three round trips before a single sponsored review.

### 1. Org-scoped grants

Sponsor every **public** repository in an installation, rather than only the repos named by numeric id.

Two new fields on `BillingFields` (`packages/core/src/types/db.ts`):

| Field | Meaning |
|---|---|
| `ossGrantScope?: 'repos' \| 'org'` | Absent = `'repos'`. Every existing #261 grant reads unchanged — no migration. |
| `ossGrantAccount?: { id, login }` | The account the grant was written for. **Informational only** — the gate never matches on it. |

`evaluateOssGrant` (`packages/billing/src/oss-grant.ts`) gains one branch:

```
no repo context                      -> no_repo_context   (unchanged)
no ossGrantExpiresAt                 -> no_grant
scope === 'repos' && repos empty     -> no_grant
expired / unparseable                -> grant_expired     (unchanged)
scope === 'repos' && id not in list  -> repo_not_granted
!repo.isPublic                       -> repo_not_public   (both scopes)
period spend >= cap                  -> cap_exceeded      (unchanged)
otherwise                            -> eligible
```

`repo_not_granted` simply never fires under org scope, so no new `OssIneligibleReason` variant is needed and `isLapsedOssGrant()` / the `oss` block copy in `block-notify.ts` are untouched.

[...]

PR #412, opened 2026-08-20
feat(billing): OSS operator commands, dashboard, and docs (#409, stage 3)

santthosh:
**Stage 3 of 3 — the capstone** for #409. Stages [1](https://github.com/mergewatch/mergewatch.ai/pull/410) and [2](https://github.com/mergewatch/mergewatch.ai/pull/411) are merged; this makes both features reachable.

## Operator commands

| Situation | Command |
|---|---|
| Installed, some repos are OSS (open core) | a repo list — unchanged |
| Installed, everything public is OSS | `--org acme-corp` |
| Not installed yet | `--preapprove acme-corp` |

Plus `--list-preapprovals`, `--inspect --org`, and `--revoke --org`.

**`--org`** resolves the org to its installation via the JWT-only `GET /orgs/{org}/installation`, falling back to the user endpoint for a personal account. It prints every public repo currently known *and* an explicit open-core warning — under org scope there is no omission to spot the way the repo-list blast radius shows one, because coverage is unbounded by design. That makes the warning the useful thing to print instead.

**`--preapprove` refuses if the org has already installed.** A pre-approval is only ever claimed on `installation.created`; that event has already fired and will never fire again, so the row would sit unclaimed forever with nobody looking at it. It points at `--org` instead.

**Flag parsing** no longer treats a following `--flag` as a value, so `--revoke --org acme` parses as revoke-targeting-an-org rather than a repo literally named `--org`. And `--org` combined with a repo list is refused outright — two different coverage models, and guessing which one was meant is the kind of silent wrong answer a grant script must never give.

## Fixes the dashboard regression org scope would have caused

`ossStatus` bailed to `null` whenever the repo list was empty — which is the *normal* state of an org-scoped grant. The billing page would have told a sponsored org it had no grant at all.

It now mirrors `evaluateOssGrant`'s coverage test, reports `scope` and `account`, and returns `repos: null` under org scope rather than a stale list the gate ignores (rendering that list would understate coverage to the maintainer). `BillingClient` shows "All / Public repositories" and "Every public repository, including ones you create later".

`ossStatus` had to be exported to be testable — it's a pure function over billing fields, and the alternative was leaving the regression this stage fixes uncovered.

## Docs

`docs/oss-program.md` has said *"there is no installation-wide mode"* since #410 merged. Rewritten: both scopes and when to pick each, the pre-approval model and its three unattended-safety properties, the login-vs-numeric-id limitation, expiry, and a mode-selection table.

## Known duplication

`scripts/grant-oss.ts` duplicates a few constants (`#PENDING-OSS`, cap, term, TTL) and the pre-approval row shape from `@mergewatch/billing` rather than importing them. This is forced, not a shortcut: root `package.json` has no dependencies and pnpm only hoists third-party packages there, so the script can import the AWS SDK and Octokit but never `@mergewatch/*`. The file already duplicated `DEFAULT_CAP_CENTS` / `DEFAULT_TERM_MONTHS` for the same reason. Both sides carry a comment, and E2E-93 lists drift as an explicit failure mode.

If you'd rather kill the duplication, the fix is adding `@mergewatch/billing` to root `devDependencies` — I left that out because it changes install topology and Amplify builds from the workspace root.

## Verification

- `pnpm run build` — 12/12
- `pnpm run test` — 21/21; **108 lambda tests** (+8 for `ossStatus`)

New `ossStatus` coverage: org-scoped grant with no repo list reported (the regression); `repos: null` rather than a stale list; expired org grant reported inactive rather than hidden; accrual counters still surfaced; and every pre-#409 path unchanged (named grant, empty list → null, no grant → null, list without expiry → null).

Arg-parsing guards exercised manually — missing `--stage`, `--org` + repo list, no target, org login shaped like a repo, and zero/negative `--cap` / `--months` / `--ttl-days` all refuse correctly.

**Two things I could not verify here**, both covered by E2E-93:
- The live AWS paths — the `mergewatch` profile's session is expired in this environment, so no `--org` / `--preapprove` / `--list-preapprovals` run touched DynamoDB.
- Dashboard lint — `packages/dashboard` has no ESLint config and `pnpm run lint` drops into an interactive setup prompt. Pre-existing; it still builds.

**E2E-93** added covering the full operator lifecycle, including the guard cases and the dashboard rendering.

## After merge

`docs/oss-program.md` is already a graduated doc, so there's no `docs/pending` move — but E2E-91/92/93 need flipping from 🚧 to ✅ and the plan doc to `Shipped`. I'll do that as a small follow-up once this lands.

### DIFF
--- a/docs/feat/20260820-01-oss-org-grants-and-preapproval.plan.md
+++ b/docs/feat/20260820-01-oss-org-grants-and-preapproval.plan.md
@@ -115,7 +115,7 @@ The SK is `account.login.toLowerCase()` — GitHub logins are case-insensitive f
 
 ### Phase 3 — operator script, dashboard, docs
 
-- [ ] **Goal:** make both features operable and visible.
+- [x] **Goal:** make both features operable and visible. **Shipped in PR #412.**
 - **Files:** `scripts/grant-oss.ts` (`--org`, `--preapprove`, `--list-preapprovals`; `--inspect` renders scope + any pending row; `--revoke` handles org grants and pending rows), `packages/lambda/src/handlers/billing.ts` (`ossStatus` currently returns `null` when the repo list is empty — line 242 — which would make every org-scoped grant invisible; add `scope` and `account` to the payload), `packages/dashboard/app/dashboard/billing/BillingClient.tsx` (render "all public repositories" instead of a repo list under org scope), `docs/oss-program.md`, `e2e/RUNBOOK.md`, `packages/billing/src/constants.ts` (`OSS_PREAPPROVAL_TTL_DAYS = 90`).
 - **Tests:** `ossStatus` returns non-null for an org-scoped grant with no repo list; still `null` with no grant at all; script arg-parsing rejects `--org` combined with a repo list.
 - **RUNBOOK:** E2E-93 — operator lifecycle: `--org`, `--preapprove`, `--list-preapprovals`, `--inspect`, `--revoke`; `--stage` guard still refuses to default.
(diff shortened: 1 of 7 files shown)


santthosh:
## Review triage

**⚠️ `scripts/grant-oss.ts:593` — empty catch swallows installation-lookup failure — AGREED, fixed in 035630e.**

Verified against the code and the finding is exactly right. The `--preapprove` probe used two nested bare catches, so any non-404 failure on both endpoints (5xx, rate limit, network blip) read as "not installed" — and the script would then write a pre-approval for an org that *is* installed. That row can never be claimed, because `installation.created` has already fired and won't fire again. Nobody would notice until someone ran `--list-preapprovals`.

That's the same silent-failure class this feature guards against everywhere else, so it deserved a real fix rather than a log line.

Both probes now go through `findInstallationByLogin`, which returns `null` **only** when GitHub answered 404 on both endpoints, and aborts on anything else naming the endpoint and status:

> Could not determine whether acme-corp has installed the App.
> GET orgs/acme-corp/installation returned 503: …
>
> Refusing to guess. Treating this as "not installed" would write a
> pre-approval that can never be claimed; treating it as "installed"
> would refuse a legitimate one.

I went slightly further than the suggested debug-log: refusing to guess is the correct failure mode here, since both wrong answers are harmful and neither is recoverable without someone noticing.

This also fixed the same weakness one level up — `resolveOrgInstallation` previously reported *every* failure, transient ones included, as "the App is not installed", which would have sent an operator to `--preapprove` during an outage. It's now a thin wrapper over the same helper.

E2E-93 extended with the transient-failure case and its failure mode.

---

**🔵 `scripts/grant-oss.ts:502` — TODO-style comment — no action.** Self-negating; the finding body says "This is not a TODO comment finding — disregard." The comment it points at documents a duplication that's forced by module resolution (workspace packages don't resolve from the repo root), which the PR body covers.

Build 12/12, typecheck 20/20, test 21/21 still green after the fix.

Issue #416, opened 2026-08-21
E2E infrastructure: dev/prod A/B on one fixtures repo, tagged selection, deterministic grading


santthosh:
Plan doc: `docs/feat/20260821-01-e2e-ab-tags-and-grading.plan.md` (lands with stage 1).

## Problem

There is no way to test a change before it reaches production.

Dev has been dark since **2026-06-25**. The dev GitHub App (id 3027556) has **zero webhook deliveries ever recorded** — not failed deliveries, zero, meaning GitHub isn't attempting delivery at all. So `mergewatch/fixtures` is reviewed only by prod, and the dev stack has never exercised a real review.

The Sonnet 4.5 model swap (#414) shipped on the strength of a hand-built `InvokeModel` call plus one manually-opened PR, because nothing better existed. That is the gap this closes.

Two further limitations compound it:

- **Suite selection is all-or-nothing in practice.** `run-suite.sh` takes positional fixture names, so running "everything a prompt change could affect" means knowing and typing the list. A full 93-fixture run is ~30 minutes and $10–15.
- **Grading needs a model.** `.claude/commands/verify-suite.md` has an LLM read each fixture's prose README as a rubric. Useful, but non-deterministic and unusable as a hard deploy gate.

## Rejected: a separate `fixtures.dev` repo

The obvious fix — a pre-production fixtures repo the dev App reviews — was considered and rejected. It makes dev and prod run **different inputs**, which is exactly what a comparison must not do, and adds a mirror to keep in sync.

**One repo, both apps** gives input parity by construction: same PR, same commit SHA, same diff, two reviews, side by side in one thread. No sync job, no drift, and it works for any repo you temporarily want to A/B.

### 1. Stage-scoped identity (stage 1, `mergewatch.ai`)

Three constants make dev and prod indistinguishable on a shared repo:

| Constant | File | Prod value — **must not change** |
|---|---|---|
| `BOT_COMMENT_MARKER` | `packages/core/src/github/client.ts:34` | `<!-- mergewatch-review -->` |
| `INLINE_BOT_COMMENT_MARKER` | `client.ts:44` | `<!-- mergewatch-inline -->` |
| `MERGEWATCH_CHECK_RUN_NAME` | `client.ts:308` | `MergeWatch Review` |

[...]

Issue #400, opened 2026-08-20
Mention-triggered reviews load .mergewatch.yml from the default branch, not the PR head — config silently inert (6 fixtures, model footer as tracer)


santthosh:
## Summary
On the 2026-08-19 Sonnet-calibration run, 8 fixture PRs whose reviews were re-triggered via `@mergewatch review` mentions (after throttle parks) all reviewed with **the base branch's config instead of their own head-ref `.mergewatch.yml`**. Six failed with their config surface inert or inverted; the model footer proves the mechanism.

## The tracer
fixtures` main now carries a `model: sonnet-4-5` pin (fixtures#585); these fixtures' overlays **replace** that file (deleting the pin) with their own config.
- **On-open reviews** (same run): head config correctly applied — [#592](https://github.com/mergewatch/fixtures/pull/592)/[#597](https://github.com/mergewatch/fixtures/pull/597)/[#632](https://github.com/mergewatch/fixtures/pull/632)/[#633](https://github.com/mergewatch/fixtures/pull/633) footers show **claude-opus-4.6** (pin gone, deploy default) and their config keys (includePatterns, maxFiles, reviewOnMention) all worked.
- **Mention-re-triggered reviews**: [#635](https://github.com/mergewatch/fixtures/pull/635)/[#636](https://github.com/mergewatch/fixtures/pull/636)/[#637](https://github.com/mergewatch/fixtures/pull/637)/[#638](https://github.com/mergewatch/fixtures/pull/638)/[#639](https://github.com/mergewatch/fixtures/pull/639)/[#641](https://github.com/mergewatch/fixtures/pull/641)/[#642](https://github.com/mergewatch/fixtures/pull/642)/[#643](https://github.com/mergewatch/fixtures/pull/643) footers ALL show **sonnet-4-5** — main's pin — despite their heads deleting it.

## Config surfaces that went inert on the mention path (matching main's config = pin only)
- **excludePatterns** (77a/77b): `api.generated.ts` reviewed, got findings + inline comments + diagram nodes, counted in "files scanned" — exclusion did no work (worked on-open in 3 prior runs).
- **ux block** (79): default wordmark header (custom commentHeader absent entirely), scanned-lines banner present despite `showWorkDone: false`, no checklist — whole block ignored (was rendered-escaped on-open).
[...]

Issue #394, opened 2026-08-19
Mermaid diagram corrupted in the review comment: quote-parity inversion entity-escapes structural syntax (PR #392)


santthosh:
## What happened

The review comment on PR #392 shipped a **broken mermaid block** — GitHub renders a parse error instead of the diagram. Observed corruption (verbatim from the comment):

```
Q -->|tracks usage"&rsqb; R&lsqb;"TokenAccumulator"&rsqb;<br/>    <br/>    S&lsqb;"Truncation retry
wrapper"] -->|wraps| A
```

Three distinct defect shapes in one diagram:

1. **Structural syntax entity-escaped**: `R&lsqb;"TokenAccumulator"&rsqb;` instead of `R["TokenAccumulator"]`, and statements glued with literal `<br/>` **outside** any quoted label — exactly the corruption class the PR #148 sanitizer (`decodeMermaidOutsideQuotes`) was built to fix, appearing on the *output* side this time.
2. **Edge labels containing raw newlines**: `-->|fallback on\nStructuredOutputUnsupportedError|` — invalid mermaid (labels cannot span lines); the label-escape pass (`escapeMermaidLabelChars`) normalizes newlines inside `"…"` node labels but edge labels in `|…|` regions aren't covered.
3. **Label line-breaks rendered literally**: `&lt;br/&gt;` inside quoted node labels — the intended `<br/>` got double-escaped, so it displays as text.

## Suspected mechanism

The sanitizer pipeline splits the diagram into alternating unquoted/quoted segments on `"` (see `decodeMermaidOutsideQuotes`, reviewer.ts ~668). A model-emitted label containing a raw newline or an unbalanced quote **inverts the parity**: from that point on, quoted regions are treated as unquoted and vice versa — the encode pass then entity-escapes brackets in structural positions (defect 1) and skips the label internals it should have protected (defect 3). The corruption in this diagram starts exactly at a quote boundary (`|tracks usage"` …), consistent with parity inversion.

## Why this matters beyond cosmetics

There is **no syntax gate** between the diagram agent's output and the posted comment: FP-D validates file *paths* in the diagram, but nothing validates that the block parses. A corrupt diagram ships to every reader of the PR.

## Suggested fixes (in value order)

[...]

Issue #390, opened 2026-08-19
Agents emit JSON as free text — replace with schema-constrained structured outputs (categorical fix for parse-failure finding loss)


santthosh:
## Severity

**High.** Free-text JSON parsing has now caused three incidents of silent finding loss (#382: 48 parse failures in one suite run including a lost unauthenticated-endpoint critical; #385: a false 5/5 on self-admitted SQLi via the same layer; residual "Unparsed agent output: 1" on PR #388's own review). Every mitigation so far (#383 balanced-object extraction, #384 disclosure, truncation retry + JSON repair in #389) reduces the frequency — none can make malformed output impossible, because the model is *writing* JSON as prose.

## The categorical fix

Stop asking the model to write JSON. Every backend we support has schema-constrained output:

| Provider | Mechanism |
|---|---|
| Anthropic direct | forced tool use (`tool_choice: {type: "tool"}`) — output validated against the tool's JSON schema |
| Bedrock (Anthropic) | same, via Converse API `toolChoice` |
| LiteLLM (OpenAI-compat) | `response_format: {type: "json_schema"}` |
| Ollama | `format: <json schema>` (structured outputs since 0.5) |

With the findings schema enforced at the API layer, "could not parse agent JSON response" becomes structurally impossible on these paths — truncation surfaces as an API-level error/stop-reason instead of corrupt text, and multi-object/prose-wrapped responses can't exist.

## Scope

1. Extend `ILLMProvider` with an optional `invokeStructured(modelId, prompt, schema, maxTokens?, sampling?)` returning a validated object (providers that can't support it omit it; callers fall back to the current text path + hardened parser).
2. Implement in all four providers (`llm-bedrock`, `llm-anthropic`, `llm-litellm`, `llm-ollama`).
3. Migrate the findings-bearing call sites in `runReviewPipeline` (6 built-in agents, custom agents, orchestrator, verifier verdicts) — schemas already exist implicitly in the prompt "Response format" blocks; make them explicit JSON Schema constants next to the prompts.
4. Keep `tryParseJson`/repair as the fallback path for providers without structured support (some LiteLLM upstreams don't implement `json_schema`).
5. RUNBOOK: retire the parse-failure disclosure expectations to "fallback path only"; add a scenario asserting zero parse failures on a structured-output run.
[...]

PR #419, opened 2026-08-22
fix(core): dismiss only our own App's stale reviews (#418)

### LINKED ISSUE #418: dismissStaleReviews dismisses ANY bot's review — cross-stage and cross-vendor
Found while running the first live dev/prod A/B on `mergewatch/fixtures` (#416, E2E-94).

## What happened

On [fixtures#707](https://github.com/mergewatch/fixtures/pull/707), both Apps reviewed the same PR. Three seconds apart:

```
15:00:36  mergewatch-ai-dev  DISMISSED  ← dev approved, then prod dismissed it
15:00:39  mergewatch         APPROVED
```

Prod ran second, called `dismissStaleReviews`, and dismissed **dev's** review.

## Root cause

`packages/core/src/github/client.ts:607`:

```ts
for (const review of reviews.data) {
  if (review.user?.type === 'Bot' && review.state !== 'DISMISSED') {
    await octokit.pulls.dismissReview({ ... });
  }
}
```

The only filter is *"was this authored by a Bot"*. Not "was this authored by MergeWatch", and not "by this stage's MergeWatch".

## Two distinct problems

**1. Cross-stage (blocks the #416 A/B).** Whichever stage reviews second dismisses the other's review. Comments and check runs are stage-scoped after #417, but formal reviews are not — so review state, one of the more interesting things to compare, is destroyed on every A/B run.

It also runs the other way: enabling the dev App on **any** repo prod also reviews means dev can dismiss prod's review. `mergewatch/kitchensink` is currently on both Apps.

**2. Cross-bot (pre-existing, and the more serious one).** This dismisses reviews from *any* bot — CopilotAI, dependabot, CodeQL, a customer's own review bot. On a customer repo, MergeWatch silently dismisses a third party's review on every re-review.

Worth noting the inline-reply path was written specifically to avoid this class of thing (`inline-reply.ts:473`):

> require BOTH that the root is bot-authored AND that it carries the `INLINE_BOT_COMMENT_MARKER` — otherwise CopilotAI, dependabot, codeql, or any other reviewer bot's threads would qualify and MergeWatch would interfere in conversations it didn't start.

`dismissStaleReviews` never got the same guard. Problem 2 predates #416 and affects production today; the A/B just surfaced it.

## Why the obvious fix is not quite trivial

Scope dismissal to reviews authored by *this* App. But identity is awkward to obtain on the review path:

- `review.user.login` is a bare slug (`mergewatch`, `mergewatch-ai-dev`) — `gh` also reports `author.is_bot: null` here, so bot-detection heuristics are unreliable (this already bit the #416 grader).
- An APPROVE review has an **empty body** by design (#132), so there is no marker to match on.
- The review agent holds an installation token, not an App JWT, so `GET /app` is not directly available to it.

Options:

1. **Resolve the App slug once per runtime** and pass it down — the Lambda already has the App ID and private key in SSM, so it can mint a JWT at cold start, read the slug, and cache it. Most correct; adds a startup dependency.
2. **Configure it** — `GITHUB_APP_SLUG` env var, defaulted per stage. Trivial, but drifts if an App is renamed.
3. **Derive from the summary comment's author** on the same PR — self-configuring and needs no new credential, but fails on the paths where no comment exists.

## Suggested behavior

Dismiss a review only when its author matches this App's login. Anything else — another stage, another vendor's bot, a human — is left alone.

That is a **behavior change for production**: MergeWatch would stop dismissing other bots' reviews. That is the fix, but it is worth stating plainly rather than shipping quietly.

## Scope note

Problem 1 blocks E2E-94's review-state assertion and `grade-run.mjs --compare`. Problem 2 is a customer-facing defect independent of #416 and probably deserves priority.

santthosh:
`Closes #418`

`dismissStaleReviews` filtered on **`review.user?.type === 'Bot'` and nothing else**, so it dismissed every bot's review on the PR.

## Two bugs in one line

**1. Cross-vendor — customer-facing today.** MergeWatch dismissed CopilotAI, dependabot, CodeQL, or a customer's own reviewer bot on every re-review. The inline-reply path guards against precisely this class of interference — `inline-reply.ts:473` names CopilotAI and dependabot explicitly — but `dismissStaleReviews` never got the same guard. This predates #416 entirely.

**2. Cross-stage — blocked the A/B.** With both Apps on one repo, whichever stage reviewed second dismissed the other's review. Observed live on `fixtures#707`:

```
15:00:36  mergewatch-ai-dev  DISMISSED   ← dev approved…
15:00:39  mergewatch         APPROVED    ← …prod dismissed it 3s later
```

It runs both ways, so enabling the dev App on any repo prod also reviews degrades prod's output. `mergewatch/kitchensink` is currently on both.

## The fix

Dismissal is scoped to reviews whose author matches our own App login.

**Identity is derived, not configured.** The review path already posts its summary comment before dismissing (`review-agent.ts:940/943` vs `:981`), so that comment's author *is* us — in every stage, in self-hosted, with no env var to set and nothing to drift when an App is renamed. One API call per process, cached.

I chose this over the three options I listed in the issue. An `GITHUB_APP_SLUG` env var was trivial but drifts silently on rename; a JWT `GET /app` lookup was most "correct" but needs a credential the review agent doesn't hold and a cold-start dependency. Deriving from a comment we demonstrably authored needs neither and self-corrects.

**Unknown identity dismisses nothing.** On the `postSummaryOnClean: false` path there's no comment, and the lookup can fail. Leaving our own review stale is cosmetic; dismissing a stranger's is a correctness and trust problem — so the ambiguous case fails toward inaction, and logs why.

**Logins compare with `[bot]` normalized.** REST and GraphQL disagree about the suffix, and `gh pr view --json reviews` returns a bare `mergewatch` with `author.is_bot: null`. Matching raw strings would silently fail to recognise our own review — that exact mismatch already bit the #416 grader, so it's pinned by a test here.

## Tests (+13, 950 core total)

`dismissStaleReviews`: dismisses our own · leaves the **other stage's** alone · leaves **CopilotAI / dependabot / CodeQL** alone · leaves humans alone · skips already-`DISMISSED` · normalizes `[bot]` in both directions · dismisses **nothing** for `undefined` / `null` / `''` identity · survives a `dismissReview` 403 without throwing.

`resolveAppLogin`: reads the login · caches to one call per process · keeps stages separate under different cache keys · returns null on a failed lookup · returns null when the comment has no author.

## Verification

build 12/12 · typecheck 20/20 · test 21/21

Extends **E2E-94** with the review-state expectation this unblocks and a third-party-bot case, plus the two matching failure modes.

## Note on rollout

This is a **behavior change in production**: MergeWatch stops dismissing other bots' reviews. That's the fix, and it's stated plainly rather than shipped quietly. Prod's deploy is currently parked at the approval gate on an earlier run, so this won't reach production until that's cleared.

### DIFF
--- a/e2e/RUNBOOK.md
+++ b/e2e/RUNBOOK.md
@@ -3270,6 +3270,8 @@ Then apply any fixture as usual (`scripts/apply-fixture.sh 01-clean-pr`).
 - [ ] Push a second commit → each stage **updates its own** comment in place; neither posts a duplicate and neither 403s trying to edit the other's
 - [ ] Click "Re-run" on the **dev** check → only dev re-reviews; prod's comment and check are untouched
 - [ ] Click "Re-run" on the **prod** check → only prod re-reviews
+- [ ] **Each stage keeps its own formal review** — prod shows `APPROVED` and dev shows `APPROVED`, neither `DISMISSED` (#418). Before that fix whichever stage reviewed second dismissed the other's review ~3s later
+- [ ] A third-party reviewer bot on the same PR (CopilotAI, dependabot, CodeQL) still has its review intact after a MergeWatch re-review (#418)
 - [ ] Inline findings carry the stage-scoped inline marker, and a 👍 on a dev inline comment is recorded against dev's stores only
 - [ ] On a repo with **only** the prod App, behavior is identical to before #416 — one comment, one check named `MergeWatch Review`
 
@@ -3278,6 +3280,8 @@ Then apply any fixture as usual (`scripts/apply-fixture.sh 01-clean-pr`).
 - ❌ A 403 in either stage's logs when updating a comment — it matched the other App's comment and tried to edit what it doesn't own
 - ❌ **Prod posts a duplicate comment on an existing PR** — the prod marker changed. This is the severe one: it orphans every bot comment in the wild, not just fixtures
 - ❌ Dev's "Re-run" button does nothing — `isMergeWatchCheckRun` still matches only the prod name, so dev ignores its own re-request
+- ❌ Either stage's formal review shows `DISMISSED` — `dismissStaleReviews` is dismissing across Apps again (#418). This also degrades prod on any repo the dev App shares with it
+- ❌ A third-party bot's review is dismissed — the #418 guard regressed, and MergeWatch is interfering with a vendor it has no business touching
 - ❌ Both stages reviewing `kitchensink` or any repo outside the A/B — the dev installation wasn't scoped down
 
 ---
--- a/packages/core/src/github/client.test.ts
+++ b/packages/core/src/github/client.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, vi } from 'vitest';
+import { describe, it, expect, vi, beforeEach } from 'vitest';
 import {
   mergeScoreToReviewEvent,
   buildInlineComments,
@@ -10,6 +10,9 @@ import {
   removePRReaction,
   submitPRReview,
   findExistingBotComment,
+  dismissStaleReviews,
+  resolveAppLogin,
+  __resetAppLoginCache,
 } from './client.js';
 import type { Octokit } from '@octokit/rest';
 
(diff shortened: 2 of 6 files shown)


santthosh:
## Review triage

**🔵 ×3 — "New TODO comment detected" (`client.ts:607`, `review-agent.ts:980`, `review-processor.ts:840`) — disagreeing, no change.**

All three flag `// #418 — …` comments. These are issue-reference comments documenting *why* the code is shaped the way it is, which is this repo's established convention rather than a task marker:

- **124** such `// #NNN —` comments already exist across `packages/`, spanning **31 distinct issues** (#193, #195, #231, #233, #235, #240, #261, #262, #310, #334, #336, #337, #350, #355, #357, #358, #359, #369, #370, #372, #380, #382, #385, #386, #387, #390, #394, #400, #409, #416).
- This diff introduces **zero** actual `TODO` / `FIXME` / `HACK` / `XXX` markers (`git diff main...HEAD | grep -E "^\+.*\b(TODO|FIXME|HACK|XXX)\b"` → 0).

The suggestion to "remove after the fix is verified in production" would be actively harmful here. These comments carry the reasoning that makes the code defensible — that `dismissStaleReviews` must never touch another vendor's review, and that unknown identity has to fail toward inaction. Deleting that once the fix is live is how the guard gets removed again by someone who doesn't know why it exists. `inline-reply.ts:473` carries the equivalent explanation for the same class of bug, and it's the reason that path never had this defect.

Worth noting the detector reached the right conclusion in its own body each time — *"This appears to be a reference comment rather than a traditional TODO"* — and reported it anyway. Same self-negating shape appeared on #411 and #412 in this repo, so it may be worth a look at that agent's prompt separately from this PR.

No code changes. Gate still green: build 12/12, typecheck 20/20, test 21/21 (950 core).

PR #422, opened 2026-08-22
feat(billing): record GitHub Marketplace purchases for attribution (#421)

### LINKED ISSUE #421: Handle GitHub Marketplace webhook (marketplace_purchase) for App publishing
Publishing the GitHub App to the Marketplace requires a **Manage webhook** section on the listing: *Payload URL*, *Content type*, *Secret*. GitHub POSTs Marketplace events (`marketplace_purchase`) there.

There is currently **no Marketplace handling anywhere in the codebase** — only GTM references in `docs/strategy/`.

## What to enter today, and what it will (not) do

| Field | Value | Notes |
|---|---|---|
| Payload URL | `https://wet81p7nzf.execute-api.us-west-2.amazonaws.com/prod/webhook` | The existing prod webhook (`WebhookUrl` output of the `mergewatch` stack) — **if** we reuse it; see the fork below |
| Content type | **`application/json`** — not negotiable | `verifySignature` (`packages/lambda/src/handlers/webhook.ts`) is HMAC-SHA256 over the **raw body**. `x-www-form-urlencoded` wraps the payload as `payload=<urlencoded>`, which breaks both `JSON.parse` and the signature semantics |
| Secret | must equal what the handler verifies against | Today that is SSM `/mergewatch/prod/github-webhook-secret`. A distinct secret needs a distinct SSM param and a route that knows to use it |

**Important:** filling this in unblocks the listing but delivers no functionality. The handler's event switch ends in

```ts
default:
  console.log(`Ignoring unhandled event type: ${githubEvent}`);
```

so a `marketplace_purchase` event is silently ignored and returns **200**. GitHub will show green deliveries while nothing is granted, revoked, or recorded. That is a safe failure — no 500s, no retry storms — but it is not a working integration, and green deliveries actively hide that.

## ✅ Decisions (2026-08-22)

**Fork 1 — reuse `/webhook`.** No new route, no new SSM param. The Marketplace listing's secret must be set to the **same value** as SSM `/mergewatch/prod/github-webhook-secret`, because one route verifies against one secret.

**Fork 2 — free plan on Marketplace; paid stays on Stripe.** The listing carries a **free plan only**. Discovery and install happen through Marketplace; monetization stays on MergeWatch's existing SaaS billing (Stripe prepaid credits, the 5-review free tier, OSS grants). GitHub never collects money on our behalf.

That removes the entire double-billing risk class: there is no plan→entitlement mapping, no Marketplace-aware branch in `billingCheck` / `recordReview`, and no path where a customer is charged by both GitHub and Stripe.

**What is left is small and well-bounded:** verify the event, record it for attribution, and never silently swallow it.

### Settled form values

| Field | Value |
|---|---|
| Payload URL | `https://wet81p7nzf.execute-api.us-west-2.amazonaws.com/prod/webhook` |
| Content type | `application/json` |
| Secret | same value as SSM `/mergewatch/prod/github-webhook-secret` |

---

## ~~Design fork 1 — one endpoint or two?~~ (settled: reuse)

**Reuse `/webhook`.** No infra change. Marketplace events are distinguished by the `X-GitHub-Event: marketplace_purchase` header, exactly like `installation` and `pull_request` are today. Requires setting the Marketplace secret to the **same value** as the App webhook secret, because one route verifies against one secret.

**Add `/marketplace`.** Its own SAM route, its own SSM secret (`/mergewatch/{stage}/marketplace-webhook-secret`), independent rotation, and Marketplace traffic never shares a code path with PR traffic. Costs a template change and a second Lambda route.

Recommendation: **reuse `/webhook`** unless we want independent secret rotation. The dispatch switch is already the established seam for event types, and a second route duplicates signature handling for no behavioral gain.

## ~~Design fork 2 — the billing collision~~ (settled: free plan only, paid stays on Stripe)

MergeWatch already has a complete billing system: Stripe prepaid credits, a 5-review lifetime free tier, and OSS grants (#409). GitHub Marketplace billing is a **second, independent source of truth** — GitHub collects the money and tells us about it.

Two very different scopes:

**(a) Free-only listing.** No paid plans. `marketplace_purchase` still fires for the free plan, so we must handle it, but there is no entitlement to grant beyond what installation already gives. Small, and it satisfies the listing requirement honestly.

**(b) Paid plans on Marketplace.** We must map plan → entitlement, honour upgrades/downgrades/cancellations, and — critically — **suppress Stripe billing for Marketplace-billed installations**, or a customer pays twice. `billingCheck` / `recordReview` would need a Marketplace-aware branch alongside the existing OSS one.

This decides the size of the work. (a) is a day; (b) is a feature with real money-correctness risk.

## Purchase can precede installation — we have already solved this shape

`marketplace_purchase` carries the purchasing **account** (login, id, type), not an installation. A user can purchase and install in either order, and GitHub does not guarantee ordering.

That is the same problem as OSS pre-approval (#409/#411): an entitlement that arrives before the installation it attaches to. The proven pattern is already in the codebase — a `#PENDING-OSS` sentinel row keyed by lowercased account login, claimed on `installation.created`, with:

- a conditional write so a webhook redelivery cannot reset an amended grant,
- the pending row marked claimed only **after** the entitlement lands,
- a claimed row inert so uninstall/reinstall never re-grants.

A `#PENDING-MARKETPLACE` row should reuse that shape rather than invent a second one. See `packages/billing/src/oss-preapproval.ts` and `docs/oss-program.md`.

Note the known limitation carries over: matching on account **login**, not numeric id, because the installation does not exist yet. A rename between purchase and install silently misses.

## Events to handle

`marketplace_purchase` actions: `purchased`, `changed`, `cancelled`, `pending_change`, `pending_change_cancelled`.

`changed` and `cancelled` carry a `previous_marketplace_purchase` for diffing. Downgrades and cancellations arrive as `pending_change` first and take effect at the end of the billing cycle — treating `pending_change` as immediate would cut off a customer who has already paid through the period. Verify the exact payload shape against GitHub's docs before relying on field names.

## Acceptance criteria (scoped to the decisions above)

- [x] Fork 1 decided — reuse `/webhook`
- [x] Fork 2 decided — free plan only, paid stays on Stripe
- [ ] `marketplace_purchase` is verified, parsed, and routed — never silently swallowed by `default:`
- [ ] Signature verified (already true for `/webhook`); an invalid signature returns 401 and is not processed
- [ ] Handling is **idempotent** — GitHub redelivers, and a redelivered `purchased` must not double-record
- [ ] The purchasing account is recorded for attribution: which installs came from Marketplace, and when
- [ ] Purchase-before-install works — a purchase arriving before `installation.created` is not lost, reusing the #409 pending-row pattern
- [ ] **Nothing is revoked on `cancelled`.** With a free plan and Stripe-side paid billing, a Marketplace cancellation must not touch entitlements — the customer may still have the App installed and credits on file. Record it, change nothing.
- [ ] `changed` / `pending_change` / `pending_change_cancelled` **log loudly**. With a free-only listing these should not occur; if one arrives it means a paid plan was added to the listing and this handler is now under-scoped
- [ ] Unknown actions log loudly rather than passing silently — a green delivery must not mean "ignored"
- [ ] Unit tests: each action, redelivery, unknown action, bad signature, purchase-before-install
- [ ] An `e2e/RUNBOOK.md` scenario
- [ ] `docs/` note covering the listing fields, the secret, and why Marketplace never bills

## Deliberately NOT built

- **No plan→entitlement mapping.** There is one plan and it is free.
- **No Marketplace branch in `billingCheck` / `recordReview`.** Stripe remains the only money path.
- **No revocation on cancel.** See above — it would be actively wrong here.

If paid Marketplace plans are ever added, all three become required at once, and `changed` arriving in the logs is the tripwire that says so.

## Out of scope

- Listing copy, pricing, and screenshots (tracked in `docs/strategy/backlog.md` as a `[gtm]` item)
- The GitHub Marketplace **API** for querying plans (`GET /marketplace_listing/...`) — only needed for reconciliation, not for the webhook

## Note

Filling in the three listing fields is safe and unblocks publishing today. Because the listing is **free-plan-only**, there is no window where a customer pays GitHub and receives nothing — the risk that made ordering matter does not exist under this decision.

The remaining exposure is smaller and purely informational: until this ships, `marketplace_purchase` is silently dropped by the `default:` branch, so Marketplace-sourced installs are indistinguishable from direct ones in our data. Attribution is lost, not access.

santthosh:
`Closes #421`

Publishing the App to the Marketplace requires a webhook on the listing. `marketplace_purchase` previously fell through the dispatch `default:` — it returned **200 and was silently dropped**, so GitHub showed green deliveries while nothing was recorded. That hides the gap rather than surfacing it.

## The decisions this implements

**Reuse `/webhook`** — Marketplace events are distinguished by the `X-GitHub-Event` header exactly as `installation` and `pull_request` are, so the listing's secret must equal the App's. A dedicated `/marketplace` route was rejected: it duplicates signature handling for no behavioral gain.

**Free plan only; paid stays on Stripe** — that removes the entire double-billing risk class. No plan→entitlement mapping, no Marketplace branch in `billingCheck`/`recordReview`, no path where someone is charged by both GitHub and Stripe. So this **grants nothing and revokes nothing** — it records the purchase for attribution.

### Listing configuration

| Field | Value |
|---|---|
| Payload URL | `https://wet81p7nzf.execute-api.us-west-2.amazonaws.com/prod/webhook` |
| Content type | **`application/json`** |
| Secret | same value as SSM `/mergewatch/prod/github-webhook-secret` |

Content type isn't a preference: `verifySignature` is HMAC-SHA256 over the **raw body**, and form-encoding wraps the payload as `payload=<urlencoded>` — every delivery would 401.

## Why storage is account-keyed

A Marketplace event carries an **account**, never an installation, and the installations table is partitioned by `installationId` — so a login can't be resolved to an installation without a scan.

Rather than scan, the `#MARKETPLACE` sentinel row **is** the record (SK = lowercased login, the same idiom as `#SETTINGS` / `#AGENTS` / `#PENDING-OSS`), and `installation.created` attaches it. For a free plan GitHub processes the purchase *before* redirecting to install, so purchase-then-install is the normal order — and account-keying means neither ordering loses it.

## Deliberate non-behaviors, each pinned by a test

**`cancelled` revokes nothing.** With a free plan and Stripe-side billing, a Marketplace cancellation says nothing about whether the customer still has the App installed or credits on file. Revoking would look like obvious symmetry and quietly break a paying customer. The test asserts zero writes to any `#SETTINGS` row.

**A redelivery never double-records and never moves `purchasedAt`** — that timestamp is the attribution signal, and a redelivery rewriting it makes the data meaningless.

**Re-attaching is a no-op**, so uninstall/reinstall doesn't flap the attribution onto a new installation id.

**Nothing propagates a failure.** GitHub surfaces failed deliveries on the listing and retries; a retry storm over an attribution record would be self-inflicted. Every path returns 200.

## The tripwire

`changed` / `pending_change` / `pending_change_cancelled` can't occur under a free-only listing. They're still recorded, and log:

```
[marketplace] UNDER-SCOPED: action=changed for <account>. The listing is
supposed to carry a free plan only — a plan change implies paid plans were
added, which needs entitlement mapping (see #421).
```

If paid plans are ever added, three things turn on at once: plan→entitlement mapping, a Marketplace branch in the billing gate, and revocation on cancel. That line is how you find out.

## Verification

build 12/12 · typecheck 20/20 · test 21/21 — **184 billing** (+16), **127 lambda** (+15)

Store: sentinel-partition keying and login normalization · idempotent redelivery (one row) · `purchasedAt` never moves · cancel records without revoking · attachment survives a later rewrite · unexpected actions recorded not dropped.

Dispatch: purchase and cancel recorded · all three unexpected actions warn `UNDER-SCOPED` · missing account ignored without throwing · recording failure still 200 · self-hosted no-ops · bad signature 401s before any work · attach fires only on `installation.created` and not on `deleted`/`suspend`/`new_permissions_accepted`.

Adds **E2E-97** and `docs/marketplace-listing.md`.

## Scope note

Self-hosted is untouched — there's no Marketplace listing for it, and the handler is behind `isSaas()`.

### DIFF
--- a/docs/marketplace-listing.md
+++ b/docs/marketplace-listing.md
@@ -44,6 +44,16 @@ A Marketplace event carries an **account** (login + numeric id), never an instal
 
 For a free plan GitHub processes the purchase **before** redirecting to install, so purchase-then-install is the normal order and the one that must work. Because the record is account-keyed and durable, neither ordering loses the purchase.
 
+### Known limitation — install-then-purchase is not attached
+
+Attachment runs on `installation.created`. If an account that **already** has the App installed later buys the free plan from the listing, the `#MARKETPLACE` record is written but never denormalized onto that installation's `#SETTINGS` row.
+
+Attribution is not lost — the record exists and is queryable by account login. Only the convenience copy is missing, so a query joining on `marketplaceAccountLogin` would miss those installations while a query over the `#MARKETPLACE` partition would not.
+
+Fixing it properly requires resolving an account login to an installation, which needs either a table scan (the installations table is partitioned by `installationId`) or an App JWT the review path does not hold. Neither is justified for attribution. Pinned by a test so the behavior is a deliberate tradeoff rather than an accident.
+
+Under a free plan GitHub processes the purchase before redirecting to install, so this is the uncommon ordering.
+
 ### Deliberate non-behaviors
 
 **`cancelled` revokes nothing.** With a free plan and Stripe-side paid billing, a Marketplace cancellation says nothing about whether the customer still has the App installed or credits on file. Revoking would look like obvious symmetry and would quietly break a paying customer.
--- a/docs/marketplace-listing.md
+++ b/docs/marketplace-listing.md
@@ -1,0 +1,69 @@
+# GitHub Marketplace listing — webhook and attribution
+
+**Status:** ✅ Shipped (#421)
+
+MergeWatch is listed on the GitHub Marketplace with a **free plan only**. Discovery and installation happen through the listing; **money never does**. Paid conversion stays on MergeWatch's own SaaS billing — Stripe prepaid credits, the 5-review free tier, and OSS grants (`docs/oss-program.md`).
+
+That single decision removes the entire double-billing risk class: there is no plan→entitlement mapping, no Marketplace branch in `billingCheck` / `recordReview`, and no path where a customer is charged by both GitHub and Stripe.
+
+## Listing configuration
+
+The listing's **Manage webhook** section takes three values:
+
+| Field | Value |
+|---|---|
+| Payload URL | the `WebhookUrl` output of the stage's stack — prod: `https://wet81p7nzf.execute-api.us-west-2.amazonaws.com/prod/webhook` |
+| Content type | **`application/json`** |
+| Secret | the same value as SSM `/mergewatch/{stage}/github-webhook-secret` |
+
+**Content type is not a preference.** `verifySignature` computes HMAC-SHA256 over the **raw request body**. `application/x-www-form-urlencoded` wraps the payload as `payload=<urlencoded>`, which breaks both `JSON.parse` and the signature comparison.
+
+**The secret must match the App's webhook secret**, because the listing reuses the App's `/webhook` endpoint and one route verifies against one secret. Marketplace events are distinguished by the `X-GitHub-Event: marketplace_purchase` header, exactly as `installation` and `pull_request` are.
+
+A dedicated `/marketplace` route with its own secret was considered and rejected: it duplicates signature handling for no behavioral gain. It becomes worth it only if Marketplace and App secrets need independent rotation.
+
+## What the handler does
+
+Records the purchase for **attribution** — which installations arrived via Marketplace, on what plan, and when. It grants nothing, because installation already grants access, and it **revokes nothing**.
+
+### Storage
+
+A Marketplace event carries an **account** (login + numeric id), never an installation, and the installations table is partitioned by `installationId` — so a login cannot be resolved to an installation without scanning. Rather than scan, the account-keyed row *is* the record, under a `#MARKETPLACE` sentinel partition (the same idiom as `#SETTINGS`, `#AGENTS`, `#PENDING-OSS`):
+
+| Field | Meaning |
+|---|---|
+| `repoFullName` | Sort key — the **lowercased** account login. Not a repository. |
+| `accountLogin` / `accountId` | GitHub's original casing, and the immutable numeric id. |
+| `planName` / `planId` | The plan at the time of the event. |
+| `lastAction` | Most recent action seen. |
+| `purchasedAt` | First `purchased` sighting. **Never overwritten** — a redelivery must not move when the account arrived. |
+| `cancelledAt` | Set on `cancelled`. Recorded only. |
+| `attachedInstallationId` / `attachedAt` | Set once `installation.created` attached the record. |
+
+`installation.created` then attaches it, writing `marketplaceAccountLogin`, `marketplaceAccountId`, `marketplacePlanName`, and `marketplaceAttachedAt` onto that installation's `#SETTINGS` row so the dashboard needs no second lookup.
+
+For a free plan GitHub processes the purchase **before** redirecting to install, so purchase-then-install is the normal order and the one that must work. Because the record is account-keyed and durable, neither ordering loses the purchase.
+
+### Deliberate non-behaviors
+
+**`cancelled` revokes nothing.** With a free plan and Stripe-side paid billing, a Marketplace cancellation says nothing about whether the customer still has the App installed or credits on file. Revoking would look like obvious symmetry and would quietly break a paying customer.
+
+**Re-attaching is a no-op.** `installation.created` is redeliverable, and an uninstall/reinstall produces a new installation id. Keeping the first attachment means the record reflects where the purchase originally landed rather than flapping.
+
+**Failures never propagate.** GitHub surfaces failed deliveries on the listing and retries; a retry storm over an attribution record would be a self-inflicted outage. Every path returns 200.
+
+**SaaS only.** Behind `isSaas()` — self-hosted has no Marketplace listing.
+
+### The tripwire
+
+`changed`, `pending_change`, and `pending_change_cancelled` should never occur under a free-only listing. If one arrives it means a **paid plan was added to the listing** and this handler is now under-scoped. It is still recorded, and it logs:
+
+```
+[marketplace] UNDER-SCOPED: action=changed for <account>. The listing is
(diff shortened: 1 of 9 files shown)


santthosh:
## Review triage

**🔵 `marketplace.ts:26` — "New TODO comment" — rejecting the framing, accepting the substance.**

**The framing is wrong.** The flagged line is prose explaining *why* the storage is account-keyed. `// #NNN —` and explanatory comments are this repo's convention — 124+ such issue-reference comments exist across `packages/`, and this diff adds **zero** `TODO`/`FIXME`/`HACK` markers. This is the fourth finding of this exact self-negating shape (also #411, #412, #419), where the body reasons its way to "this is documentation" and reports it anyway.

**The buried suggestion was right, though**, and I've acted on it:

> Consider adding explicit tests or documentation for the install-then-purchase edge case if it can occur.

It can occur, and it was neither tested nor documented. Attach runs only on `installation.created`, so an account that **already** has the App installed and later buys the free plan from the listing gets a `#MARKETPLACE` record that is never denormalized onto its `#SETTINGS` row.

Worth being precise about the impact: **attribution is not lost.** The record exists and is queryable by account login — only the convenience copy is missing, so a query joining on `marketplaceAccountLogin` misses those installations while one over the `#MARKETPLACE` partition does not.

Fixing it properly needs login → installation resolution, which requires either a table scan (the installations table is partitioned by `installationId`) or an App JWT the review path doesn't hold. Neither is justified for an attribution record, and under a free plan GitHub processes the purchase *before* redirecting to install — so this is the uncommon ordering, not the normal one.

Pushed in `docs(billing): pin install-then-purchase as a known limitation`: a test asserting the record exists, is unattached, and writes nothing to any installation row, plus a "Known limitation" section in `docs/marketplace-listing.md`. It now reads as a deliberate tradeoff rather than an oversight.

Gate after the change: build 12/12, typecheck 20/20, test 21/21 — **185 billing** (+1).

Issue #424, opened 2026-08-23
Context architecture: retrieval over a git checkout, agents call tools


santthosh:
Tracking issue for the context architecture. **The design record is `docs/feat/20260822-01-context-architecture.design.md`** (branch `docs/context-architecture-design`) — read that first; this issue is the tracker, not the design.

## Why

Grew out of #423 (large PRs hard-fail with `Input is too long for requested model`). Two data points moved the conversation past that bug:

1. **#423's reported failure is not a context-size problem.** 80% of `orca#117`'s 711KB diff is a single `tsconfig.tsbuildinfo`, which default `excludePatterns` does not cover. Without it the diff is ~35K tokens and fits Sonnet 4.5's 200K comfortably.
2. **Full codebase + git history are planned.** At that point available context is permanently ~1000× any window (a 10MB repo is ~2.5M tokens; history is larger). No model size fixes it.

So the question stops being *"what do I cut to fit?"* and becomes *"what do I go get?"* — retrieval, not budgeting.

## Direction

**A git checkout is the corpus, and agents call tools against it.**

Git is already a content-addressed store and `git clone --filter=blob:none` is already lazy retrieval — an earlier plan to build a symbol index keyed by blob SHA, with DynamoDB and Postgres implementations, was reimplementing git. The diff becomes the *query*, not the context.

```
corpus       git checkout at PR head (blobless partial clone)
                ↓
selection    the diff is the QUERY  ·  rg / git log -L / git blame / tree-sitter
                ↓
assembly     stable prefix ┃ cache breakpoint ┃ volatile suffix
                ↓
top-up       bounded tool calls
                ↓
provenance   record what was actually in context
```

## Settled

- Git checkout as corpus; blobless partial clone; **per PR head, not repo head** (a PR branches from `main@X` while `main` moved to `Y` — retrieval against the wrong tree gives *plausible* wrong answers)
- Structural queries, not embeddings — exact, deterministic, free, air-gapped. Embeddings are a SaaS-only additive layer if ever; core retrieval must never depend on them or the deployments diverge in quality
- **Agents call tools; they are not handed a payload**
- Async, **queued not blocking** — the SQS path from #355 already parks and retries
- **Self-hosted parity required** (air-gapped is supported — it is why the Ollama provider exists). This ruled out hosted embeddings, managed vector stores, and S3-only artifacts, and is what made the checkout approach obviously right
- Self-hosted needs a **disk volume**, not blob storage
[...]

PR #426, opened 2026-08-23
fix(core): bound review input so oversized diffs skip, not hard-fail (#423)

### LINKED ISSUE #423: Large PRs hard-fail with "Input is too long for requested model" — context regression from #414
Large PRs now fail their review outright with:

> **Review failed** — MergeWatch encountered an error: Input is too long for requested model.

Observed on [santthosh/orca#117](https://github.com/santthosh/orca/pull/117) and #116, prod, 2026-08-22.

## This is a regression from #414

#414 changed the default review model from `us.anthropic.claude-opus-4-6-v1` to `us.anthropic.claude-sonnet-4-5-20250929-v1:0`. Those models do **not** have the same context window:

| Model | Context |
|---|---|
| Opus 4.6 (previous default) | **1M tokens** |
| Sonnet 4.5 (current default) | **200K tokens** |

A 5× reduction. There is **no input-size guard anywhere in the review path** — the diff is passed through unbounded — so Opus 4.6's 1M window had been silently absorbing very large diffs. Sonnet 4.5's 200K cannot, and the failure is hard rather than graceful.

I sized the swap on cost and verified invocability, sampling params, and structured output. I did not check whether any path depended on the larger context window. That was the gap.

## Evidence

```
16:16:32  Starting review for santthosh/orca#117
16:16:35  [model] santthosh/orca#117 using us.anthropic.claude-sonnet-4-5-20250929-v1:0 (source=deploy-default)
16:16:36  WARN  LLM invocation failed during agentic file fetching, falling back to no-context analysis: ValidationException: Input is too long for requested model.
16:16:36  WARN  [structured] structured invocation failed — falling back to the text path: ValidationException: Input is too long for requested model.
16:16:36  ERROR Review failed for santthosh/orca#117: ValidationException: Input is too long for requested model.
```

`source=deploy-default` — orca has no `.mergewatch.yml`, so it takes the deployment default. **Every repo without a model pin is exposed.**

Diff sizes on the same repo, same hour:

| PR | Diff | Result |
|---|---|---|
| #115 | 12.5 KB (~3K tokens) | ✅ reviewed, 0 findings |
| #116 | (GitHub refused to generate the diff) | ❌ failed |
| #117 | **711,765 bytes ≈ 178K tokens** | ❌ failed |

178K of diff plus the system prompt, conventions, and the `maxTokensPerAgent` output reservation exceeds 200K. It fit comfortably in 1M.

## Every fallback collapses

The failure is not for want of fallbacks — all of them hit the same wall:

1. Agentic file fetching → too long → falls back to no-context analysis
2. Structured invocation (#390) → too long → falls back to the text path
3. Text path → too long → **hard failure**

The WARN pairs repeat per agent, so all 8 parallel agents fail identically. The user sees "Review failed", not a degraded review — which is the worst available outcome: no findings, no partial result, no actionable guidance.

## Existing limits do not cover this

- `maxFiles: 50` — #117 has 31 files, well under.
- `maxContextKB: 256` — governs **related-file context**, not the diff itself.

Neither bounds diff size. A single 700KB diff sails past both.

## Options

**A. Add an input guard (needed regardless of model choice).** Estimate prompt tokens before invoking; when over budget, either truncate the diff to the highest-value files or skip with an explicit "diff too large to review (N tokens vs M limit)" check-run message. A review that says *why* it can't proceed is far better than `Review failed`, and the skip path already exists for docs-only and maxFiles.

**B. Move the default to a 1M-context model.** `us.anthropic.claude-sonnet-4-6` is **already priced** in `packages/core/src/llm/pricing.ts` (`$3.30/$16.50` on the `us.` profile) and already handled by `acceptsSamplingParams`. Per the model table Sonnet 4.6 carries a **1M context window at Sonnet-tier pricing** — that restores what the swap silently lost while keeping the ~40% cost saving over Opus 4.6. Needs the same live-invoke verification #414 did (plain + forced-tool paths) before flipping `infra/params/*.env`.

**C. Revert to Opus 4.6.** Restores the 1M window immediately; gives back the cost saving.

**D. Sonnet 4.5's 1M-context beta.** Requires a beta header; Bedrock availability unverified. Not worth pursuing over (B).

Recommendation: **A + B.** (B) removes the immediate breakage and is a strictly better model than what we shipped; (A) is the durable fix, because any model has a limit and hard-failing on a large PR is wrong at any context size.

## Acceptance criteria

- [ ] A PR whose diff exceeds the model's context does **not** hard-fail; it either reviews a bounded subset or posts an explicit, actionable skip
- [ ] The check-run message names the actual cause and size, not `Input is too long for requested model`
- [ ] The bound is derived from the configured model rather than hard-coded, so a future model change cannot silently reintroduce this
- [ ] `orca#117` (711KB diff) produces a useful outcome
- [ ] Unit tests: under budget reviews normally; over budget degrades; the boundary case
- [ ] An `e2e/RUNBOOK.md` scenario with a deliberately oversized fixture
- [ ] If the default model changes: live-invoke verified (plain + forced-tool) before `infra/params/*.env` is touched, per the rule in that file

## Blast radius

Any repo on the deploy default with a large PR, since #414 deployed (2026-08-22 ~15:30 UTC). Small-PR repos are unaffected — the freeCodeCamp cohort's 53 reviews all completed normally. Repos pinning `model:` in `.mergewatch.yml` are unaffected.

santthosh:
`Closes #423` · architecture context in #424 · pairs with #425 (Sonnet 4.6)

The review path had **no input-size guard at all** — the diff went to the model unbounded. Whichever context window the configured model happened to have was the only thing between a large PR and a failure. That held while the default was a 1M model; when #414 moved it to a 200K one, large PRs started failing with a raw `ValidationException: Input is too long for requested model` after every fallback collapsed. No findings, no partial result, no guidance.

## Three layers, cheapest first

**1. Default `excludePatterns` gained build artifacts** — `*.tsbuildinfo`, `*.map`, `*.snap`, `__generated__/**`, `__snapshots__/**`, `generated/**`, `*.gen.*`, `*.pb.go`, `vendor/**`, `coverage/**`, `.next/**`, `*.wasm`.

**This alone fixes the reported bug.** 80% of `orca#117`'s 711KB diff was one `tsconfig.tsbuildinfo`; without it the diff is ~35K tokens and fits even the 200K model that failed it. The PR was never too big — we were reviewing a build cache.

**2. `maxFileDiffKB` (default 128)** drops any single file whose diff section exceeds it, regardless of pattern. A list only catches artifacts we've already been bitten by; this catches the *next* one, because a hand-written source file with a 128KB+ diff essentially doesn't exist.

Reported **separately** from pattern exclusions — a pattern match is the operator's intent, a size drop is ours, and conflating them makes it impossible to tell why a file vanished from review.

**3. A pre-flight input budget** from the resolved model's context window (new `packages/core/src/llm/context-windows.ts`, mirroring the shape and home of the pricing table). Over budget → a neutral **"Review skipped — diff too large"** check naming the estimated tokens, the budget, the model, and the remedy.

## Two deliberate choices

**An unknown model assumes the *smaller* window.** The two ways to be wrong aren't symmetric: too small produces a visible, actionable skip; too large reproduces exactly the opaque failure this replaces. Self-hosted runs whatever `LLM_MODEL` names, so it's the deployment most likely to hit this — and it gets the safe side.

**The budget subtracts `maxTokensPerAgent`**, because output shares the window. A request whose input plus `max_tokens` exceeds the context is rejected even when the input alone would have fit.

Also includes a **"the configured default model has a known context window"** guard, mirroring the pricing guard from #414 — a default that silently fell back to the assumed window would halve everyone's usable budget without anyone noticing.

## ⚠️ This is a bound, not a solution

A large enough diff overflows any window. #425 raises the cliff 5× and this one makes hitting it survivable — but the durable answer is retrieval over a checkout rather than shipping the whole diff, which is #424.

## Verification

build 12/12 · typecheck 20/20 · test 21/21 — **984 core** (+47), 121 server

The budget tests use the real numbers from the incident: the 711,765-byte diff rejected on Sonnet 4.5, accepted on Sonnet 4.6, and accepted on Sonnet 4.5 once the 571KB artifact is removed — which is the clearest statement of what actually went wrong.

Also covered: unknown model assumes 200K · `undefined` model · output reservation shrinks the budget · never a negative budget · assumed-window flag · size drop vs pattern exclusion reported distinctly · zero/negative cap means *no cap*, not drop-everything · a file exactly at the cap is kept · every new default pattern matches · ordinary source files unaffected.

Server test mocks updated for the new `filterDiff` return shape — the contract changed, so the mock should match rather than the runtime defending against a stale one.

Adds **E2E-98**.

### DIFF
--- a/docs-site/configuration/mergewatch-yml.mdx
+++ b/docs-site/configuration/mergewatch-yml.mdx
@@ -453,3 +453,17 @@ MergeWatch reads the file from the default branch, not from the PR's head branch
 <Warning>
 If you rename your default branch, MergeWatch automatically detects the new default on the next webhook. No action is required.
 </Warning>
+
+## `maxFileDiffKB`
+
+Maximum size, in KB, of a single file's diff section. Default `128`. A file whose diff exceeds this is dropped from review even if no `excludePatterns` entry matches it.
+
+`excludePatterns` only catches artifacts you have already met. This catches the next one: a hand-written source file with a 128KB+ diff essentially does not exist, so a section that large is generated, vendored, or minified.
+
+Dropped files are always reported in the logs, distinctly from pattern exclusions — a pattern match is your intent, a size drop is MergeWatch's.
+
+```yaml
+maxFileDiffKB: 128   # set 0 to disable the size rule entirely
+```
+
+Related: MergeWatch also refuses to review a diff that would exceed the configured model's context window, posting a neutral **"Review skipped — diff too large"** check that names the estimated size, the budget, and the model. Previously such a PR failed with a raw provider error.
--- a/e2e/RUNBOOK.md
+++ b/e2e/RUNBOOK.md
@@ -227,6 +227,7 @@ A **`—`** in the Tags column means the scenario has **no fixture directory** a
 | [E2E-87](#e2e-87-337--date-only-range-bounds-include-their-whole-day) | `/api/analytics` date-only `start_date`/`end_date` expand to the UTC day's edge instants at the boundary (`end_date=2026-08-16` includes the whole 16th); full timestamps pass through untouched; identical on both backends; UTC bucketing documented in the route (#337) | 2m | 30s | #337 | `rollup` |
 | [E2E-88](#e2e-88-355--pr-burst-resilience) | A PR burst never silently loses reviews: throttles park the review (`pending` + in_progress "rate limited" check, never terminal FAILURE) and admission control paces the backlog — SQS `MaximumConcurrency` on SaaS, Postgres `SKIP LOCKED` worker at `REVIEW_CONCURRENCY` on self-hosted; exhaustion lands in a DLQ/`status='dead'`, visibly (#355) | 20m | n/a | #355 | — |
 | [E2E-95](#e2e-95-416--selective-suite-runs-by-tag-mode-or-changed-paths) | `TAGS`/`MODE` on every fixture; `--tag` / `--mode` / `--changed-files` resolve a subset; unmapped paths and unknown tags fail loudly rather than silently running nothing (#416) | 2m | n/a | fixtures#705 | `tooling` |
+| [E2E-98](#e2e-98-423--oversized-diffs-skip-with-a-reason-never-hard-fail) | Build artifacts excluded by default and oversized files dropped by size; a diff still over the model's context budget yields a neutral "diff too large" skip naming sizes and remedy, never a raw ValidationException (#423) | 4m | 60s | #426 | `tooling`, `skip` |
 | [E2E-97](#e2e-97-421--marketplace-purchase-recorded-and-attached) | A Marketplace `purchased` is recorded under `#MARKETPLACE` and attached to the installation on `installation.created`; redelivery does not double-record or move `purchasedAt`; `cancelled` revokes nothing (#421) | 5m | 60s | #422 | `marketplace` |
 | [E2E-96](#e2e-96-416--deterministic-grading-of-a-suite-run) | `grade-run.mjs` asserts a run against `expect.json` with no model, exits 1 on regression; UNGRADED is never PASS; `--compare` reports dev/prod divergence (#416) | 3m | n/a | fixtures#706 | `tooling` |
 | [E2E-94](#e2e-94-416--dev-and-prod-review-the-same-pr-without-colliding) | With both Apps installed on one repo, a single PR gets two independent reviews — separate comment markers, separate check runs, each stage updating only its own and re-running only from its own button; prod's identity is byte-identical to pre-#416 (#416) | 5m | 2m | #416 | — |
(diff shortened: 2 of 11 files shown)


santthosh:
## Review triage

**🔵 `defaults.ts:226` — "New TODO comment: build artifacts exclusion list" — no change.**

The finding's own suggestion says *"No action required"*, and I agree. The flagged text is the comment explaining **why** the pattern list is deliberately incomplete and how `maxFileDiffKB` covers the gap — the reasoning that stops someone later "fixing" the list by trying to enumerate every artifact, or deleting `maxFileDiffKB` as redundant.

For the record, this is the sixth finding of this shape in this repo (#411, #412, #419, #422 ×2, now #426). `// #NNN —` and explanatory comments are the established convention — 124+ across `packages/` spanning 30+ issues — and this diff adds zero real `TODO`/`FIXME`/`HACK` markers. The detector reliably reasons its way to "this is documentation" in its own body and reports it anyway.

That's not a blocker here, but it's a consistent enough false-positive shape that the agent's prompt seems worth a look independently of any one PR.

No code changes. Gate unchanged: build 12/12, typecheck 20/20, test 21/21 (984 core).

PR #474, opened 2026-08-25
fix(core): bound every review write path so an oversized comment truncates

### LINKED ISSUE #468: Comment body size guard: an oversized review comment fails the POST and vanishes silently
## Why

**There is no comment body length guard anywhere in the codebase.** Verified: no length check, truncation, or size handling on any write path in `packages/core/src/github/client.ts`, and no occurrence of the limit anywhere in `packages/core/src`.

GitHub caps issue-comment bodies at **65,536 characters**. An oversized body fails the POST with a `422 Validation Failed`. The failure mode is the worst available one: the review ran, spent real tokens, produced real findings — and the developer sees *nothing on the PR at all*. Not a truncated comment, not an error comment. Silence, indistinguishable from MergeWatch not being installed.

Nothing reaches the limit routinely today, which is why this hasn't fired. But several sections have no bound on their contribution:

- Findings sections — `maxFindings` caps the count, not `description` + `suggestion` length per finding
- `📎 Previously reported findings` — grows with every carried-over finding across a long-lived PR
- W10's "Related concerns clustered into this finding" blocks, which preserve every absorbed sibling's framing
- The mermaid diagram, which is model-generated and unbounded
- Custom agent findings (#385), which bypass every filtering layer by design

A PR with a large diagram, a long delta history, and verbose findings is a plausible path to 65KB with nothing malfunctioning. #401's runaway agent (1,430 near-empty findings from one response) shows the abnormal path.

Should land before #467, which adds per-finding evidence to the comment.

## Scope — four write paths, not one

| Path | Limit | Current behaviour | In scope |
|---|---|---|---|
| `postReviewComment` / `updateReviewComment` (`client.ts:193`, `:223`) | 65,536 | 422 → review vanishes | ✅ primary |
| `createStandaloneReviewComment` (`client.ts:459`) + `submitPRReview` `comments[]` | 65,536 per comment | inline body built at `client.ts:545` from raw model prose, unbounded | ✅ |
| `createCheckRun` `output.summary` (`client.ts:338`) | 65,535 | wrapped in try/catch, **explicitly non-fatal by design** — fails silently today | ✅ |
| `submitPRReview` `body` | 65,536 | one-line pointer, never at risk | ❌ |

The inline path matters as much as the primary one: it posts `title` + `description` + `suggestion` straight through with no cap, and a fix that only touches the issue comment would look complete while leaving it exposed.

### 1. Section-aware assembly in `formatReviewComment`

**This is the design decision the rest depends on.** `formatReviewComment` currently pushes into one flat `lines: string[]` across 84 call sites and returns `lines.join('\n')` (`comment-formatter.ts:661`). There is no section structure to drop from.

Re-invoking the formatter with existing flags disabled does **not** work: `showDiagram` / `showSummary` / `showIssuesTable` exist, but there is no flag for the `📎 Previously reported findings` block (driven by `delta`), none for the work-done section, and none for the per-severity sub-sections — info and warnings both render off `findings`. Shedding would jump straight from "everything" to "no findings at all."

So: refactor the assembly into an ordered `sections: Array<{ id, priority, text }>`, joined by a budget-aware step. Contained to one file; the existing tests assert on the output string, so most should survive unchanged.

### 2. Budget and drop order

`MAX_COMMENT_BODY = 65_536`, working budget 60,000. **The guard runs on the final marked body** — `postReviewComment` prepends the marker after the formatter returns (`client.ts:201`), so a budget applied to the formatter's output alone is measuring the wrong string.

Drop order, lowest priority first: diagram → work-done → previously-reported → info → warnings. The merge verdict, summary, and all criticals are **never** droppable; if those alone exceed budget, truncate within criticals and say so.

Priority order matters more than the exact budget: whatever survives is what MergeWatch is asserting about the PR, and that must never be a partial findings list that reads as complete.

### 3. Truncation notice

`_This comment was truncated to fit GitHub's size limit. [View the full review](url)._`

`reviewDetailUrl` is **optional** in `FormatOptions` and absent on self-hosted deployments with no dashboard. Degrade to `_N section(s) omitted to fit GitHub's comment size limit._` — never render a dead link or a notice implying a destination that doesn't exist.

### 4. Inline comment bodies

Cap each inline body at build time (`client.ts:545`) by truncating `description` / `suggestion` with an ellipsis, same `PREV_FINDING_FIELD_MAX` precedent used elsewhere. The fingerprint HTML comment must survive truncation — it's load-bearing for re-review matching.

### 5. Check run summary

Truncate `output.summary` to its own cap before the call. Note the existing catch block makes this path fail silently *by design* ("don't fail the review if the check run fails"), so a size failure here is invisible today — truncating is the only way it stays visible.

### 6. Defensive final assertion

A last check on every assembled body before POST, so a future section that grows unbounded cannot resurrect the silent-failure path.

## Acceptance criteria

- [ ] A comment body exceeding the budget posts successfully, truncated, rather than failing with 422
- [ ] The guard measures the marked body, not the formatter's raw output
- [ ] Drop order is tested: verdict, summary, and all criticals survive a body forced well over budget
- [ ] The findings list is never silently partial — if findings are dropped, the notice says how many
- [ ] The truncation notice degrades correctly when `reviewDetailUrl` is absent
- [ ] Inline comment bodies are capped, and the `mw-fp` fingerprint survives truncation
- [ ] Check run `output.summary` is truncated before the call
- [ ] A pre-POST assertion guarantees every body is under the hard cap regardless of how it was assembled
- [ ] `updateReviewComment` (re-review path) is covered, not just create

santthosh:
Closes #468.

There is no comment body length guard anywhere on the write paths today. GitHub caps an issue-comment body at 65,536 characters and rejects anything larger with a `422` — so an oversized review fails the POST and the developer sees **nothing on the PR at all**. Not a truncated comment, not an error comment. Silence, indistinguishable from MergeWatch never having run, after the review already spent the tokens.

## Four paths, each bounded

| Path | Bound | How |
|---|---|---|
| `postReviewComment` / `updateReviewComment` | 65,536 | section shedding + hard assertion on the **marked** body |
| `buildInlineComments` | per-field | `title` / `description` / `suggestion` capped; `mw-fp` preserved |
| `createCheckRun` `output.summary` | 65,535 | truncated before the call |
| `submitPRReview` `body` | — | one-line pointer, never at risk (out of scope per the issue) |

## The design decision

`formatReviewComment` pushed into one flat `lines: string[]` across 84 call sites, so there was no structure to drop from — and re-invoking with the existing flags does not help: there is no flag for the `📎 Previously reported findings` block, none for work-done, and info and warnings both render off `findings`. Shedding would have jumped straight from *everything* to *no findings at all*.

So the assembly now builds ordered sections. Drop order: **diagram → work-done → previously-reported → info → warnings**. The verdict, the summary and **every critical** are never droppable. If criticals alone exceed budget they truncate in place and say so.

> Whatever survives is what MergeWatch is asserting about the PR. It must never be a partial findings list that reads as complete.

**Where the guard measures.** The formatter cannot know the marker — it is stage-dependent (`reviewMarker(stage)`) and prepended *after* the formatter returns. So shedding runs against a 60,000 budget in the formatter, and the hard cap is asserted in the client on `markedBody`, the actual final string. That is the literal requirement from the issue, and it doubles as the defensive backstop for any future section that grows without a budget of its own.

## Notes on the smaller calls

- **The notice degrades.** With no `reviewDetailUrl` — self-hosted, no dashboard — it drops to `N section(s) omitted…` and renders no link. A notice implying a destination that does not exist is its own defect.
- **It names counts** (`60 info findings`), not just "some sections", so the surviving list cannot read as complete.
- **It renders above the findings.** A reader who learns content is missing only after scrolling past them has already drawn the wrong conclusion.
- **The fingerprint outlives the prose.** `mw-fp` is appended *after* truncation — `/resolve` rebuilds the suppression key from it (#182), so cutting it would silently break re-review matching.
- **The check run truncation is the visible-failure fix.** That path's `try/catch` is non-fatal by design, so an oversized summary fails invisibly today.

## Verification

The section refactor preserves the exact push order, so a normal review is **byte-identical**. All **68 pre-existing formatter assertions pass untouched** — that is the back-compat proof, not a new test I wrote to agree with myself.

```
build      12/12
typecheck  20/20
test       21/21   (1,061 tests — 21 new)
```

New tests cover: drop order (diagram sheds first, info survives), criticals/summary/verdict surviving a body forced well over budget, the marked-body measurement, notice counts, notice degradation without a URL, notice placement, fingerprint survival, title-cap vs `extractInlineCommentTitle`, check run truncation, and byte-identity for normal-sized bodies.

RUNBOOK gains **E2E-100**, marked unit-test-gated: nothing reaches 65KB routinely, which is exactly why this never fired, so the unit tests are the honest gate rather than a fixture that would pass without exercising anything.


santthosh:
Fixed in `67ee395`'s follow-up — thanks, this was a real bug and the one field the PR left unbounded.

One correction to the mechanism, because it makes the failure **quieter** than reported rather than louder: `body.slice(0, negative)` does not return an empty string. A negative end index counts from the end, so `'abcdef'.slice(0, -2)` is `'abcd'`. The body therefore stays long, the fingerprint is appended anyway, and the assembled comment exceeds the cap — the exact silent 422 this PR exists to prevent, reintroduced by the guard itself. Verified: `node -e "'abcdef'.slice(0,-2)"` → `abcd`.

`Math.max(0, …)` alone would clamp the index but still emit `body + fingerprint` over the cap whenever the fingerprint is itself larger than the limit. So the fix bounds the fingerprint first: if it cannot leave `INLINE_MIN_PROSE` (1,000 chars) of room for the finding it annotates, it is dropped with a warning. Losing `/resolve` matching on one pathological finding beats losing the comment whole. The slice index is then non-negative by construction — `Math.max` stays as a local guarantee rather than an inferred one.

Two tests added: an 80,000-char fingerprint now yields a body within the cap with no `mw-fp`, and a large-but-usable fingerprint still round-trips intact.

santthosh:
Both remaining findings check out as false positives. Verified each against the code rather than taking or dismissing them at face value.

**⚠️ `client.ts:650` — "truncation ignores the marker prefix length".** It does not. `marker` is the *first component* of `body`:

```ts
let body = `${marker}\n**🔴 ${title}**\n\n${description}…`;
if (body.length + fingerprint.length > MAX_COMMENT_BODY) { … }
```

So `body.length` already counts the marker, and the check measures the assembled string — not the fields before prepending. The invariant holds unconditionally: when the branch is not taken the body is under the cap by definition; when it is, `slice(0, MAX - fp - 1)` + `…` + fingerprint lands at exactly `MAX` in the worst case.

Rather than leave that as an argument in a comment, I pinned it: a new parameterised test runs the longest stage markers (`dev` at 30 chars, prod at 26, plus an over-long suffix) with every field maxed and asserts `≤ MAX_COMMENT_BODY`. A future marker change now fails a test instead of eroding a claim.

**🔵 `comment-formatter.ts:401` — "`Math.max(0, …)` guard is absent".** It is present, and is exactly the line the suggestion asks for:

```ts
return `${body.slice(0, Math.max(0, budget - notice.length))}${notice}`;
```

No change needed.

Worth noting the prior round's finding was **real and I fixed it** — the unbounded fingerprint — so this is not a blanket dismissal. `build 12/12 · typecheck 20/20 · test 21/21`, now 26 new tests.

santthosh:
On the unverified critical (`client.ts:640`) — W2 was right not to confirm it. The finding's own three worked examples each arrive at **exactly 65,536**, which is *at* the cap, not over, and it then concludes "causing the assembled body to exceed the cap" without producing a case that does.

The bound is structural. When the truncation branch is taken:

```
slice(0, MAX - F - 1)  →  MAX - F - 1
        + '…'          →  MAX - F        ('…' is U+2026, one JS char)
        + fingerprint  →  MAX            for every F the guard admits
```

Rather than argue arithmetic on an unverified finding, I measured it. Swept raw fingerprint lengths 48,000–49,000 (the region where `marked.length` crosses `MAX_COMMENT_BODY - INLINE_MIN_PROSE`):

```
peak assembled body: 65536
cap:                 65536  → never exceeded
```

Committed as a test over the boundary region, so the invariant is pinned rather than asserted.

**Why no extra clamp.** The suggestion offers a final backstop check. I'm declining that one specifically because it would be unreachable code guarding a case the arithmetic rules out — and this repo's bar is explicit about not handling cases that cannot occur. The genuine version of that requirement (acceptance criterion 8) is the pre-POST assertion in `postReviewComment`/`updateReviewComment`, which exists and covers assembly the formatter cannot see.

**One thing I am deliberately not fixing here**, flagged as a follow-up rather than silently ignored: all of this counts UTF-16 code units via `.length`. If GitHub's 65,536 limit is bytes rather than characters, a body of emoji and box-drawing characters could be up to ~3× over while measuring as under. The issue framed the limit as characters and every test asserts `.length`, so changing the unit is out of scope for this PR — but it is worth its own issue.

Scoreboard for this review round: 1 real finding (the unbounded fingerprint — fixed), 3 false positives (marker length, missing `Math.max`, this boundary case). `build 12/12 · typecheck 20/20 · test 21/21`, 27 new tests.

santthosh:
Fixed — and this one was a fair catch, though not quite for the stated reason.

**The claim rests on "GitHub's API likely counts bytes."** I can't settle that from here: GitHub's own validation error reads `Body is too long (maximum is 65536 characters)`, and #468 specifies characters. But "likely bytes" is not something to build a guarantee on either, and I had raised the same doubt myself in the previous comment.

**So I made the question moot rather than answering it.** The real defect was narrower than described: the inline path was the one place assembling to *exactly* the cap with zero slack — the summary-comment path has always worked against 60,000 of a 65,536 budget. That asymmetry was mine and it was worth removing.

`INLINE_BODY_MAX = MAX_COMMENT_BODY - 512`, plus the same treatment for the check-run summary. Measured across raw fingerprints 48,000–49,000:

```
peak chars: 65024 | peak UTF-8 bytes: 65030 | cap: 65536
under cap by chars: true | by bytes: true
```

Holds under **either** interpretation, which is the honest way to close a question I can't definitively resolve. The boundary test now asserts strictly-less-than so the reserve can't be quietly removed later.

I did **not** take the suggestion to replace `…` with `...`. That trades a rendering detail for arithmetic that a reserve solves properly, and it would leave every other multi-byte character — including the 🔴 in the title, which is 4 bytes to 2 code units — still unaccounted for. A margin covers the whole class; swapping one glyph covers one instance of it.

**Clustered warning — open redirect on `reviewDetailUrl`: not changed, and not a vector.** It is built by us:

```ts
`${DASHBOARD_BASE_URL}/dashboard/reviews/${encodeURIComponent(`${repoFullName}:${prNumberCommitSha}`)}`
```

The dynamic half is percent-encoded, the base is operator configuration, and an operator who can set `DASHBOARD_BASE_URL` already controls the deployment. The line also predates this PR — the footer has always rendered it.

Running total across four review rounds: **2 real findings fixed** (unbounded fingerprint, zero-margin inline ceiling), **4 declined with reasons** (marker length, missing `Math.max`, the exact-boundary case, this URL warning).

santthosh:
Declining this one — it conflates the two ceilings, and the evidence is direct.

```ts
export const MAX_COMMENT_BODY = 65_536;              // client.ts:71  — GitHub's hard cap
const INLINE_BODY_MAX = MAX_COMMENT_BODY - 512;      // client.ts:92  — our working ceiling = 65,024
```

The finding's own reasoning says the total "becomes exactly `INLINE_BODY_MAX`" — that is **65,024**, which is 512 *below* `MAX_COMMENT_BODY`. So the title, "can produce a body that exceeds MAX_COMMENT_BODY", is contradicted by the finding's own arithmetic.

It also states that landing there "violates" the `toBeLessThan(MAX_COMMENT_BODY)` assertion. That assertion is green right now:

```
$ npx vitest run src/github/client.test.ts -t 'boundary'
Tests  1 passed | 105 skipped (106)
```

65,024 < 65,536. The invariant it describes as violated is the one the reserve established last round.

The suggestion — reserve one *more* character so the result is strictly below the working ceiling — would change nothing about safety. `- 1` versus `- 2` against a self-imposed ceiling that already sits 512 under the hard cap is arbitrary, and taking it would encode a wrong premise into the code.

**On the loop.** This is the third round on the same few lines, and each framing has been different and each incorrect: first that the marker was uncounted, then that the exact-boundary case exceeded the cap, now that landing on the working ceiling exceeds the hard cap. The earlier two rounds produced **two genuine fixes** I was glad to make — the unbounded fingerprint and the zero-margin ceiling — so this is not a dismissal of the reviewer. But the remaining objection has been answered with measurements three times, and I am not going to keep pushing commits against arithmetic that is checkable and checked.

**Open-redirect warning:** unchanged, and answered in full above — `reviewDetailUrl` is `${DASHBOARD_BASE_URL}/dashboard/reviews/${encodeURIComponent(…)}`, operator-configured base, percent-encoded tail, and the line predates this PR.

Final state: `build 12/12 · typecheck 20/20 · test 21/21` (1,067 tests, 27 new). Peak assembled inline body across the boundary sweep: **65,024 characters / 65,030 UTF-8 bytes** against a 65,536 cap — under it by either counting scheme.

santthosh:
**Merging over the red review check**, with the override recorded here rather than left implicit.

The blocking finding states the assembled body "exceeds `MAX_COMMENT_BODY`" because it lands on `INLINE_BODY_MAX`. Those constants differ by the 512-character reserve added in the previous round:

| | value |
|---|---|
| `MAX_COMMENT_BODY` (hard cap) | 65,536 |
| `INLINE_BODY_MAX` (working ceiling) | 65,024 |
| peak assembled body, measured | 65,024 chars / 65,030 UTF-8 bytes |

The finding also reports a `toBeLessThan(MAX_COMMENT_BODY)` assertion as violated; that test is green. `Build & Test` is green and all 1,067 tests pass.

This is a deliberate human decision to merge, not an unnoticed red check. Two findings from earlier rounds on this same PR were real and were fixed — the unbounded fingerprint and the zero-margin inline ceiling — so the override applies to this finding specifically, not to the reviewer's judgement generally.

The byte-vs-character question underneath all of this is being filed separately: it is a legitimate open question about how GitHub counts the 65,536 limit, and it deserves its own investigation rather than being settled inside a PR about something else.

Issue #473, opened 2026-08-25
Orchestrator rationale: make the biggest black box explain its drops


santthosh:
Stage 5 of #467. **Gated** — do not implement until #469–#472 have produced real data.

## Why

The orchestrator is the largest single source of dropped findings and the only stage in the pipeline that cannot explain itself for free. `runOrchestratorAgent` takes N tagged findings and returns M, with no record of which N died or why. After #470, its drops record as `stage: 'orchestrator'` with no reason — attributable but not explained.

## Why it's gated

Making it explain itself costs output tokens on **every review**, not just ones where someone asks. And given #385's history — the orchestrator's anti-pedantry rule classified a blocking org agent's findings as pedantry, which meant the #235 blocking gate could never fire — there is real risk in asking a model to justify decisions it has already made. A post-hoc rationale that reads well is not evidence that the decision was right, and shipping one could make a bad filter *more* trusted rather than less.

So this needs evidence before it's worth paying for. #469–#472 supply it: once the dashboard shows the trail, we can measure how often "the orchestrator dropped it and we don't know why" is actually the answer a developer hits. If it's rare, the cost isn't justified. If it's the most common unexplained gap, it is.

## What (provisional)

Extend `ORCHESTRATOR_SCHEMA` (`agents/schemas.ts:49`) with:

```ts
dropped: {
  type: 'array',
  items: {
    type: 'object',
    properties: {
      file: { type: 'string' },
      line: { type: 'integer' },
      title: { type: 'string' },
      reason: { type: 'string', description: 'One sentence' },
    },
    required: ['title', 'reason'],
  },
}
```

Feed those reasons into the `FindingOutcome` records for `stage: 'orchestrator'`.

## Open questions to answer with #469–#472 data

- How often do developers hit an unexplained orchestrator drop?
- What does the extra output cost per review, measured rather than estimated?
- Does the model's stated reason match what actually happened, or does it confabulate? Needs a fixture set where the true reason is known.
- Should this be opt-in per repo rather than default-on?

## Acceptance criteria

_To be written when the gate opens. Do not implement from this description._

santthosh:
## Gate evidence — first measurement from live trace data

#469–#472 are all in production as of today, so the gate's inputs now exist. First read, from **66 traces / 188 outcomes** across the dev and prod trace tables.

### Open question 1 — how often is "the orchestrator dropped it and we don't know why" the answer?

| stage | non-surfaced outcomes | explains itself? |
[...]

Issue #469, opened 2026-08-25
Evidence: per-finding proof in the PR comment


santthosh:
Stage 1 of #467. Depends on #468 (comment size guard) landing first.

## Why

A developer reading a finding on their PR has no reason to believe it: a title, a description, a suggestion, and a confidence percentage — all model prose. Nothing shows the code the finding is about, which agent raised it, or that a second model checked it against the full file and confirmed it.

The W2/FP-E verifier already generates exactly that confirmation — one sentence citing specific code — for every critical and warning. It formats it into a `console.warn` and throws it away (`reviewer.ts:2075`). We also fetch the full file at head for every finding's path (`fetchFindingFileContents`) and keep none of the cited lines. This is plumbing, not new inference.

### Plumbing

`groundingFileContents` is local to `runReviewPipeline` and is **not** on `ReviewPipelineResult`, so evidence must be attached to each finding inside the pipeline rather than reconstructed by the caller. The verifier's `reason` is likewise discarded inside `verifyFindings`, whose return type carries only the `verification` tag.

Threading it out means extending three parallel finding types: `OrchestratedFinding`, `Finding` (`comment-formatter.ts:14`), and `ReviewFinding` (`types/db.ts:434`). `reviews.findings` is `jsonb` in Postgres (`storage-postgres/src/schema.ts:49`), so **no migration is needed**.

### Evidence must be short

A finding that needs a paragraph to justify itself has failed to justify itself. Three elements, nothing else:

1. **Cited code** — the anchor line ±1, max 3 lines, from `groundingFileContents`. Highest-value element by a wide margin: it makes the claim checkable at a glance instead of taking the model's word for it.
2. **The verifier's reason** — one sentence, already specified as "One sentence citing the specific code" in `VERIFIER_VERDICT_SCHEMA`. Hard-capped at 200 chars, matching the `PREV_FINDING_FIELD_MAX` precedent (`reviewer.ts:1283`).
3. **Agent convergence** — rendered *only* when two or more agents independently raised it. "security + bugs agreed" is real signal; naming a single agent restates the category and is noise.

Explicitly cut from the comment, dashboard-only (#467 Stage 4): the grounding result ("anchor confirmed" is internal jargon nobody can act on) and confidence-vs-floor (the percentage already renders via `showConfidence`; appending the floor is chrome).

Budget: ~350 characters per critical, keeping comment growth bounded and predictable.

### Rendering

[...]

Issue #467, opened 2026-08-25
[Tracker] Explainability: finding evidence and the filter decision trail


santthosh:
Tracker. Implementation lives in the child issues below.

## Why

MergeWatch blocks merges. A gate that can't explain its verdict can't be trusted, can't be overridden intelligently, and can't be audited — and right now ours explains almost nothing.

Two gaps, two audiences:

**A developer reading a finding has no reason to believe it.** Title, description, suggestion, confidence percentage — all model prose. Nothing shows the code the finding is about, which agent raised it, or that a second model checked it against the full file. The W2/FP-E verifier generates exactly that confirmation for every critical and warning, formats it into a `console.warn`, and throws it away (`reviewer.ts:2075`).

**Nobody can find out what was filtered, or why.** Twelve stages in `runReviewPipeline` can delete, merge, or demote a finding. Every one logs to `console.warn` and nowhere else. What survives is one scalar: `suppressedCount: Math.max(0, totalRawFindings - filteredFindings.length)`.

#401 already showed a bare count is actively misleading — two production reviews reported "1430" and "3869 findings removed by dedup & quality filters" on 4-line diffs, a malfunctioning agent's junk described as productive filtering. The fix was to stop counting junk. The deeper problem — a count cannot distinguish good filtering from a malfunction — is still here.

**Most of the evidence already exists and is discarded**: the verifier's per-finding `reason`, `groundingFileContents` (we fetch every cited file and keep none of the lines), and FP-C's cross-agent convergence signal — the strongest confidence signal in the system, destroyed by the merge that detects it. This is mostly plumbing, not new inference.

`ReviewDetail.tsx` ends with `{/* Future: reasoning flow, agent traces, etc. */}`.

## Design split

| | PR comment | Dashboard |
|---|---|---|
| **Evidence for a surfaced finding** | ✅ inline, terse | ✅ full |
| **Filtered findings + reasoning** | ❌ unchanged | ✅ full |

Evidence attaches to findings the developer is already reading. The filtered set — high volume, read once if ever — lives in the dashboard. This keeps comment growth bounded by `maxFindings` rather than by the unbounded dropped-set.

## Children

| | Issue | Depends on | Status |
|---|---|---|---|
| **1** | #469 — Evidence: per-finding proof in the PR comment | #468 | ✅ shipped (#476) — see #477 |
| **2** | #470 — Filter outcome ledger | — (but see note) | ready |
[...]

Issue #401, opened 2026-08-20
Suppressed-findings counter shows implausible values (1430, 3869) on 4-line diffs


santthosh:
## Summary
Two reviews on the 2026-08-19 Sonnet run rendered absurd suppression tallies:
- [fixtures#625](https://github.com/mergewatch/fixtures/pull/625): "**Suppressed** | 1430 findings removed by dedup & quality filters" — on a 4-line diff.
- [fixtures#628](https://github.com/mergewatch/fixtures/pull/628): "**Suppressed** | 3869 findings removed" — also a ~4-line diff.

Every other review in the run shows plausible counts (1–20). Either the counter is double-accumulating (e.g. summing across retries/attempts — both PRs were in the burst window), or an agent went into runaway generation and the pipeline paid for thousands of discarded findings. Both readings deserve a look: one is a display bug, the other a token-burn bug. The review outcomes themselves were correct (625: 5/5; 628: 3/5 with the expected warning).

🤖 Filed by /verify-suite (e2e-cycle)

PR #478, opened 2026-08-25
feat(core): filter outcome ledger — record why every finding was dropped

### LINKED ISSUE #470: Filter outcome ledger: record why every finding was dropped, merged, or demoted
Stage 2 of #467. Independent of #469, but **both edit `agents/reviewer.ts` heavily — land them sequentially, not in parallel**, or they will conflict.

## Why

Twelve stages in `runReviewPipeline` can delete, merge, or demote a finding. Every one announces its decision to `console.warn` and nowhere else. What survives is a single scalar:

```ts
suppressedCount: Math.max(0, totalRawFindings - filteredFindings.length)
```

rendered as `| **Suppressed** | 13 findings removed by dedup & quality filters |`. No way to learn which thirteen, why, or whether the one thing you were worried about was among them.

We already learned a bare count is not just uninformative but actively misleading: #401 shipped because two production reviews reported "1430" and "3869 findings removed by dedup & quality filters" on 4-line diffs — a malfunctioning agent's junk, faithfully described as productive filtering. The fix was to stop counting junk. The deeper problem — a count cannot distinguish good filtering from a malfunction — remains.

This ticket captures the record. It ships no UI; #471 (persistence) and #472 (dashboard) consume it.

## What

```ts
export type FilterStage =
  | 'fp-c-line-dedup'       // cross-agent same (file,line) merge
  | 'orchestrator'          // LLM dedup / rank / maxFindings cap
  | 'w10-clustering'        // same-region consolidation
  | 'confidence-floor'      // FP-A
  | 'min-severity'          // #310
  | 'w11-scope-awareness'   // no-test-harness collapse
  | 'grounding'             // anchor / identifier / no-op-suggestion
  | 'fp-i-already-applied'  // suggestion byte-equals existing code
  | 'finding-verify'        // W2 / FP-E verifier verdict
  | 'line-proximity'        // ±3 changed-line filter
  | 'custom-agent-dedup'    // #385 re-entry
  | 'triage-suppressed';    // W3

export interface FindingOutcome {
  /** W9 fingerprint when available, else `file::T::<title>`. */
  key: string;
  file: string;
  line: number;
  severity: 'critical' | 'warning' | 'info';
  confidence?: number;
  title: string;
  /** Every agent that independently emitted this, pre-FP-C merge. */
  agents: string[];
  outcome: 'surfaced' | 'merged' | 'dropped' | 'demoted';
  stage?: FilterStage;
  reason?: string;
  /** For `merged`: the key it folded into. */
  mergedInto?: string;
}
```

> ⚠️ **Name collision — do not reuse `FindingDisposition`.** `FindingDispositionRecord` already exists at `types/db.ts:654` with entirely different semantics: a *cross-review lifetime counter* record (`surfaceCount`, `disputeCount`, `verifiedCount`, `agreementCount`, per-UTC-day buckets) backing the FB-E insights rollup, with its own store at `storage-dynamo/src/finding-disposition-store.ts`. This record is *per-review, per-finding*. Extending the existing type instead of adding a new one would corrupt the insights rollup.

`demoted` is a distinct outcome, not a flavour of dropped — grounding and the verifier both demote criticals to `unverified` rather than deleting (#385), and that decision is currently invisible.

A `TraceRecorder` threads through the pipeline; every site that today `console.warn`s a decision also records one. Reasons come from what each stage already knows — the verifier's `reason`, the confidence value against the active floor, the sibling key for a cluster merge.

**`suppressedCount` becomes derived**, not subtracted: `outcomes.filter(o => o.outcome !== 'surfaced').length`. The subtraction is measured upstream of where #385's custom-agent findings re-enter; deriving it also lets a self-consistency test catch any stage that forgets to record.

**Out of scope:** unparsed and degenerate agent responses. A response that didn't parse yields no finding object, and `isUsableFinding` rejects entries with no title. These stay as the existing `parseFailureCount` / `degenerateResponseCount` reliability warnings, which already say the right thing.

**Not in this ticket:** the orchestrator cannot explain itself without a schema change and extra output tokens (#473). Until then its drops record as `stage: 'orchestrator'` with no reason — attributable but not explained.

## Acceptance criteria

- [ ] All 12 stages record a `FindingOutcome` for each finding they drop, merge, or demote
- [ ] `surfaced` findings are recorded too, not only the removed ones
- [ ] `suppressedCount` is derived from the ledger
- [ ] A test asserts derived `suppressedCount` equals the old subtraction on a fixture where no custom agents run
- [ ] A test asserts every finding entering the pipeline has exactly one terminal `FindingOutcome`
- [ ] `demoted` is distinguishable from `dropped` in the record
- [ ] Parse failures and degenerate responses do NOT become ledger entries
- [ ] Orchestrator drops record with `stage: 'orchestrator'` and no `reason`
- [ ] No behaviour change to what the PR comment renders

## Deployment parity

Core-only and in-memory; no storage, no runtime-specific work. Both handlers already read `result.suppressedCount` (`review-agent.ts:939`, `server/src/review-processor.ts:793`), so switching it from a subtraction to a derived count reaches both for free.

- [ ] No new code in `packages/lambda` or `packages/server`
- [ ] The derived `suppressedCount` renders identically in the PR comment under both runtimes

santthosh:
Closes #470. Stage 2 of #467, landed after #469 as the issue requires — both edit `agents/reviewer.ts` heavily.

Twelve stages can delete, merge, or demote a finding. Every one announced its decision to `console.warn` and nowhere else; what reached the reader was one scalar — *"13 findings removed by dedup & quality filters"*. No way to learn which thirteen, why, or whether the thing you were worried about was among them.

That a bare count is actively misleading is not hypothetical: **#401 shipped because two production reviews reported `1430` and `3869` findings removed on 4-line diffs** — a malfunctioning agent's junk, faithfully described as productive filtering. The fix then was to stop counting junk. The deeper problem is that a count cannot distinguish good filtering from a malfunction at all.

## The design problem the ticket did not name

**Identity is not stable across the pipeline.** FP-C's cross-agent merge and W10's clustering both rewrite a *surviving* finding's title (`— and N related concerns`), so a title-derived key changes mid-flight. Recording only the absorbed siblings would leave the survivor looking like a different finding downstream, and "exactly one terminal outcome per entrant" would split into an **orphan row and a ghost row**.

Both merge stages now report `{ primaryBefore, primaryAfter, absorbed }`, and `TraceRecorder.alias()` resolves later decisions back to the row that entered. There is a test that renames twice — FP-C, then W10 — because that is the case a single-alias implementation would silently get wrong.

## Decisions worth stating

- **`demoted` is its own outcome**, not a flavour of dropped. Grounding and the verifier demote criticals to `unverified` rather than deleting (#385); that was invisible before.
- **First verdict wins.** A later stage cannot rewrite an earlier one's record, so a double-record is a no-op rather than a silent revision of history.
- **An unadjudicated entry surfaces as `dropped` with no stage** — a wiring gap shows up instead of quietly balancing the books.
- **The orchestrator is attributable, not explained.** It is an LLM returning a new set, so its drops are knowable only by difference, and it cannot explain itself without a schema change and extra output tokens (#473). A title it reworded reads as a drop plus a new entry — that is what is *observable*, and inventing a link would be a guess.
- **Parse failures and degenerate entries never enter.** A response that did not parse yields no finding object; `isUsableFinding` rejects title-less entries. Both stay reliability warnings, which already say the right thing. Two tests pin this.

## One genuinely user-visible consequence

`suppressedCount` is now **derived**, and the dead `totalRawFindings` subtraction is deleted. The old subtraction was measured *upstream* of where #385's custom-agent findings re-enter, so it counted them as suppressed.

**On a repo with custom agents the rendered number may legitimately drop.** That is the fix, not a regression — and on a repo without custom agents the two must agree exactly, which is asserted directly. RUNBOOK E2E-99 now records this so a future reader does not file it as a bug.

## Naming

Deliberately **not** an extension of `FindingDispositionRecord` (`types/db.ts:654`), which is a *cross-review lifetime counter* backing the FB-E insights rollup. Extending it would corrupt that rollup. This record is per-review, per-finding.

## Verification

```
build      12/12
typecheck  20/20
test       21/21   (1,113 tests — 19 new)
```

Core-only, as the issue's parity section requires: no new code in `packages/lambda` or `packages/server`. Both handlers already read `result.suppressedCount`, so the switch reaches both runtimes unchanged.

### DIFF
--- a/e2e/RUNBOOK.md
+++ b/e2e/RUNBOOK.md
@@ -3469,6 +3469,7 @@ Two production reviews reported `1430` and `3869 findings removed by dedup & qua
 - [ ] Where an agent does malfunction, the details drawer shows **⚠️ Malformed agent output — N agent response(s) returned unusable findings**, not an inflated Suppressed line
 - [ ] `[findings] DEGENERATE agent response: N of M entries unusable` appears in the logs when it happens
 - [ ] A normal review is unchanged — no new warning, suppressed counts still reflect real dedup work
+- [ ] **(#470)** The Suppressed number is now *derived* from the filter ledger rather than `raw − final`. On a repo with **custom agents** the number may legitimately drop: the old subtraction was measured upstream of where #385's custom-agent findings re-enter, so it counted them as suppressed. A change there is the fix, not a regression — but on a repo with no custom agents the two must agree exactly.
 - [ ] `⚠️ Unparsed agent output` (#382) still appears independently for genuinely unparseable responses
 
 **Failure modes.**
--- a/packages/core/src/agents/reviewer.test.ts
+++ b/packages/core/src/agents/reviewer.test.ts
@@ -3869,3 +3869,147 @@ describe('normalizeEvidenceReason (#469)', () => {
     expect(normalizeEvidenceReason('Line 15 interpolates id.')).toBe('Line 15 interpolates id.');
   });
 });
+
+// ---------------------------------------------------------------------------
+// Filter outcome ledger (#470)
+// ---------------------------------------------------------------------------
+
+describe('filter outcome ledger (#470)', () => {
+  const allAgents = {
+    security: true, bugs: true, style: true, summary: true, diagram: true,
+    errorHandling: true, testCoverage: true, commentAccuracy: true,
+  };
+  const empty = JSON.stringify({ findings: [] });
+  const summaryResponse = JSON.stringify({ summary: 'Refactor.' });
+  const diagramResponse = '%% overview\nflowchart TD\n  A-->B';
+
+  // Two findings on changed lines, deliberately sharing no significant token
+  // so neither FP-C (same line) nor W10 (shared wording) merges them — this
+  // isolates the stage under test.
+  const twoFindings = JSON.stringify({
+    findings: [
+      { file: 'foo.ts', line: 2, severity: 'warning', category: 'bug', title: 'Unbounded retry loop', description: 'd', suggestion: 's' },
+      { file: 'foo.ts', line: 3, severity: 'warning', category: 'bug', title: 'Missing null check on payload', description: 'd', suggestion: 's' },
+    ],
+  });
+
+  async function run(responses: string[], extra: Record<string, unknown> = {}) {
+    const llm = createMockLLM(responses);
+    return runReviewPipeline(
+      {
+        diff: sampleDiff, context: sampleContext,
+        modelId: 'heavy-model', lightModelId: 'light-model',
+        maxFindings: 25, enabledAgents: allAgents, ...extra,
+      },
+      { llm },
+    );
+  }
+
+  it('derives suppressedCount equal to the old raw-minus-final subtraction', async () => {
+    // No custom agents run here, which is the condition under which the old
+    // subtraction was correct — it double-counted #385 re-entrants otherwise.
+    const orchestrator = JSON.stringify({
+      findings: [{ file: 'foo.ts', line: 3, severity: 'warning', category: 'bug', title: 'Missing null check on payload', description: 'd', suggestion: 's' }],
+      mergeScore: 3, mergeScoreReason: 'One warning.',
+    });
+    const result = await run([
+      twoFindings, empty, empty, empty, empty, empty,
+      summaryResponse, diagramResponse, orchestrator,
+    ]);
+
+    const totalRaw = 2;                       // what the old counter measured
+    const oldSubtraction = totalRaw - result.findings.length;
+    expect(result.suppressedCount).toBe(oldSubtraction);
+    expect(result.suppressedCount).toBe(1);
+  });
+
+  it('gives every finding that entered exactly one terminal outcome', async () => {
+    const orchestrator = JSON.stringify({
+      findings: [{ file: 'foo.ts', line: 3, severity: 'warning', category: 'bug', title: 'Missing null check on payload', description: 'd', suggestion: 's' }],
+      mergeScore: 3, mergeScoreReason: 'One warning.',
+    });
+    const result = await run([
+      twoFindings, empty, empty, empty, empty, empty,
+      summaryResponse, diagramResponse, orchestrator,
+    ]);
+
+    const keys = result.filterOutcomes.map((o) => o.key);
+    expect(new Set(keys).size).toBe(keys.length);          // no finding recorded twice
+    expect(result.filterOutcomes).toHaveLength(2);          // no finding lost
+    for (const o of result.filterOutcomes) {
+      expect(['surfaced', 'merged', 'dropped', 'demoted']).toContain(o.outcome);
+    }
+    expect(result.filterOutcomes.filter((o) => o.outcome === 'surfaced')).toHaveLength(1);
+  });
+
+  it('attributes an orchestrator drop to its stage, with no reason', async () => {
+    // #473 would let the orchestrator explain itself; until then its drops are
+    // attributable but not explained, and the ledger should not invent a reason.
+    const orchestrator = JSON.stringify({ findings: [], mergeScore: 5, mergeScoreReason: 'Clean.' });
+    const result = await run([
+      twoFindings, empty, empty, empty, empty, empty,
+      summaryResponse, diagramResponse, orchestrator,
+    ]);
+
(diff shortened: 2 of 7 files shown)


santthosh:
Good catch on the asymmetry — the two keys genuinely can differ, and it was not commented. But the suggested direction inverts it: reading back the **resolved** key throws on exactly the path the fallback exists to handle.

The keys differ only when an alias points at a row that **never entered** — an orchestrator rewrite that a merge stage then renamed:

| case | `key` (resolved) | `enter()` writes under | `get(outcomeKey(f))` | `get(key)` — the suggestion |
|---|---|---|---|---|
| no alias | `= outcomeKey(f)` | `outcomeKey(f)` | ✅ found | ✅ identical |
| alias → row that never entered | dangling | `outcomeKey(f)` | ✅ found | ❌ `undefined` |

`enter()` keys off the finding itself, so reading back the key **it** used is what makes the fallback total. The resolved key has no row there, and `!` would be asserting on `undefined`.

Rather than argue it, I applied the suggestion verbatim and ran the suite:

```
FAIL  handles an alias pointing at a row that never entered
TypeError: Cannot read properties of undefined (reading 'outcome')

FAIL  does not create a duplicate row when the same finding is recorded twice
TypeError: Cannot read properties of undefined (reading 'outcome')
```

Then reverted — 15 passed.

**On the duplicate-entry half of the claim:** no duplicate arises. A second `record()` through a dangling alias resolves to the missing key, calls `enter()`, which finds the existing row under `outcomeKey(f)` and only adds the agent — then first-verdict-wins returns. There is now a test asserting exactly that, including that the *first* verdict and reason survive.

Three regression tests added, plus a comment at the call site so the asymmetry reads as deliberate rather than as the oversight it was reasonably taken for. The ordinary case — an alias whose target **did** enter — is pinned too, since that is the one this fallback is not involved in at all.

`typecheck 20/20 · test 21/21`, 22 new tests on this PR.

santthosh:
Three concerns this round. One holds and is fixed; two do not, with evidence.

## ✅ Taken — alias resolution termination

`resolve()` walked the chain with an arbitrary cap of 8. The cap prevents a hang but returns whichever key the loop happened to land on, so a cyclic chain yields a *different wrong answer per bound*. Now it breaks on the first repeated key — deterministic and independent of chain length. Two tests: a forced cycle terminates without throwing, and a three-link chain still resolves to the original row.

The pipeline only renames forward, so no cycle can occur today. Hardening anyway, because the recorder should not depend on its one caller behaving.

## ❌ Not taken — the duplicate-row claim, and especially its fix

**The suggested fix creates the very duplicate it aims to prevent.** It stores the same entry object under both the resolved and unresolved keys, and `outcomes()` iterates `entries.values()` — so one row is emitted twice. Applied verbatim:

```
FAIL  handles an alias pointing at a row that never entered
AssertionError: expected [ { …(11) }, { …(11) } ] to have a length of 1 but got 2

FAIL  does not create a duplicate row when the same finding is recorded twice
AssertionError: expected [ { …(11) }, { …(11) } ] to have a length of 1 but got 2
```

**And the scenario cannot arise.** The claim is that a second `record()` triggers "a second `enter()` call, creating a duplicate row before the first-verdict-wins guard can fire". But `enter()` is idempotent — it looks up `outcomeKey(f)` and, finding the existing row, only adds the agent:

```ts
enter(f, agent) {
  const key = outcomeKey(f);
  const existing = this.entries.get(key);
  if (existing) { existing.agents.add(agent); return; }   // ← no new row
  …
}
```

So the second `record()` lands on the same row and first-verdict-wins returns. That exact sequence is already asserted by `does not create a duplicate row when the same finding is recorded twice through a dangling alias`, which passes.

## ❌ Not taken — the "#473" comments as untracked TODOs

Issue-anchored comments are this repo's convention for recording *why* something is the way it is, not stray TODOs:

| file | issue-referencing comments |
|---|---|
| `agents/reviewer.ts` | 88 |
| `agents/reviewer.test.ts` | 94 |
| `comment-formatter.ts` | 22 |

#473 is a real tracked issue, and the comment explains why orchestrator drops carry a stage but no reason — which is precisely the kind of thing a reader of that test needs. The suggested alternative, an `excludePatterns` entry, configures which files get reviewed and has no bearing on this.

`typecheck 20/20 · test 21/21` — 24 new tests on this PR.

Issue #472, opened 2026-08-25
Dashboard: render findings, per-finding evidence, and the filtered-findings trail


santthosh:
Stage 4 of #467. Part A depends on nothing and can land immediately. Parts B and C depend on #471.

## Why

The review detail page shows model, commit, PR link, timestamps, and a settings card. That's it. The last line of `ReviewDetail.tsx` is, verbatim:

```tsx
{/* Future: reasoning flow, agent traces, etc. */}
```

This is that.

### Part A — render what already exists (no dependencies)

`ReviewDetail.tsx` renders **no findings, no summary, and no merge score at all**. Not because the data is missing — `/api/reviews/[id]/route.ts` already returns all three — but because `app/dashboard/reviews/[id]/page.tsx:56` doesn't pass them to the component.

`ReviewDrawer.tsx` (the list-view drawer) already renders findings with severity dots and confidence badges and can be borrowed from.

This is a standalone gap worth fixing on its own; it does not need to wait for the ledger.

### Part B — per-finding evidence panel (depends on #471)

Everything #469 puts in the PR comment, uncollapsed, plus the two elements deliberately cut from the comment as too jargon-heavy for that surface:

- Cited code, with a wider window than the comment's 3 lines
- Verifier verdict and reason
- Originating agents, with convergence called out
- **Confidence vs. the active floor** — `82% (floor 75)`
- **Grounding result** — anchor confirmed, identifier present

### Part C — the filtered-findings trail (depends on #471)

Every `FindingOutcome`, grouped by `FilterStage`, ordered so judgment-bearing stages (`finding-verify`, `orchestrator`, `grounding`) come before mechanical ones (`confidence-floor`, `min-severity`). Per-stage counts, each entry's reason, and `merged` entries linking to the finding they folded into.

Plus a funnel view: raw findings → each stage's in/out → surfaced.

**Orchestrator drops must display as "not explained", not as a blank reason.** Until #473 they are attributable but unexplained, and an empty cell reads as coverage — which is the exact failure mode this whole effort exists to remove.

`demoted` must render distinctly from `dropped`. A demoted critical is still a live concern the developer should read; a dropped one is not.

## Notes

The PR comment already links here via `[View full details](reviewDetailUrl)`, so no new navigation is needed.

[...]

Issue #295, opened 2026-08-14
Compliance/audit evidence export: per-merge validation record


santthosh:
## Why
Nobody exports a clean, auditable per-merge validation record (what was checked, by which signals, verdict, who overrode). This is the "system of record" surface for SOC 2 / ISO / EU AI Act change management — high value for the compliance buyer and a natural extension of the blocking gate + provenance stack. Gap #22, Tier 2 in `docs/strategy/feature-gap.md`.

## What
- Export (CSV/JSON, dashboard + API): per merged PR — provenance classification, agents/scanners run, findings + dispositions, score, check verdict, blocking events, overrides ("shipped anyway"), timestamps, config/policy snapshot.
- Builds on the reviews table, deterministic-signal records (#281), and disposition capture (#284).
- Date-range + repo filters; documented schema; stable IDs for audit referencing.

## Acceptance criteria
- [ ] Export for a repo/date range reproduces the full validation story for each merge
- [ ] Available in both SaaS and self-hosted
- [ ] Schema documented for auditors

PR #480, opened 2026-08-25
feat(storage): persist the filter outcome ledger (Dynamo + Postgres)

### LINKED ISSUE #471: Persist the filter outcome ledger (Dynamo + Postgres parity)
Stage 3 of #467. Depends on #470 (the ledger must exist before it can be persisted).

## Why

#470 builds the `FindingOutcome` ledger in memory. It dies with the Lambda invocation unless it is written somewhere. The dashboard (#472) and the future audit export (#295) both need it durable.

### ⚠️ A separate table — NOT a sort-key row on the reviews table

The obvious design — a `{prNumber}#{sha}#TRACE` sort key alongside the review — **is a correctness bug**, not just a style preference.

`queryByPR` uses `begins_with(prNumberCommitSha, :pr)` (`storage-dynamo/src/review-store.ts:99`) and all six call sites pass `${prNumber}#` as the prefix:

- `lambda/src/handlers/review-agent.ts:154`, `:236`, `:733`
- `server/src/review-processor.ts:88`, `:165`, `:557`
- `core/src/insights/disposition-writer.ts:398`

A row keyed `42#abc123#TRACE` matches that prefix by construction and would be returned as if it were a review. Worse, those calls pass `limit: 5` — so trace rows can **evict real reviews** from the previous-review lookup that feeds `previousFindings`, the delta, and FP-B. That is a silent correctness regression inside the review pipeline, not a query annoyance. Any suffix on the same sort key collides the same way.

**Dynamo:** a new `mergewatch-review-traces-{stage}` table in `infra/template.yaml` — PK `repoFullName`, SK `prNumberCommitSha`, TTL attribute. Handles the 400KB item-cap concern as a side effect.

**Postgres:** a new `review_traces` table in `storage-postgres/src/schema.ts`. Per CLAUDE.md, the generated migration SQL **must be hand-edited to `IF NOT EXISTS` / `ADD COLUMN IF NOT EXISTS`** before `pnpm run migrations:check` will pass.

Note: adding evidence fields to findings needed no migration because `reviews.findings` is `jsonb` (`schema.ts:49`). This ticket is the first that does need one.

### TTL and size

TTL shorter than the review's 90 days — this is a debugging and trust artifact, not a system of record (that's #295's job). Suggest 30 days; confirm before implementing.

Cap the retained outcome count per review with an explicit overflow marker. A #401-shaped runaway agent must not write an unbounded trace, and a truncated trace must say it was truncated rather than reading as complete.

### Interface

Extend `IReviewStore`, or add an `IReviewTraceStore` alongside it — implementer's call, but both backends must reach parity in this ticket. Self-hosted cannot ship a dashboard feature the Postgres path can't serve.

## Acceptance criteria

- [ ] Traces persist in both DynamoDB and Postgres
- [ ] `queryByPR` returns exactly the same rows before and after this change — a regression test asserts trace rows never appear in its results
- [ ] The Postgres migration passes `migrations:check` and is idempotent (`IF NOT EXISTS`)
- [ ] TTL is set and documented
- [ ] Per-review outcome count is capped, and truncation is explicitly marked in the record
- [ ] A review that fails to write its trace still posts its comment — trace persistence is never allowed to fail a review

## Deployment parity — env var checklist

If the trace table name is configurable rather than derived, the new var must land in **all** of:

- `infra/template.yaml` (SaaS table + IAM grant on both the review Lambda and the Amplify SSR role)
- `docker-compose.yml` (self-hosted; the `mergewatch` service, and the `dashboard` service if it reads traces directly)
- `.env.example`
- `packages/dashboard/next.config.js` `env` block — **an env var absent from this block does not exist at runtime on Amplify.** There is already a comment at `next.config.js:132` recording that exact failure from #315.

Prefer deriving the name from the existing stage suffix to avoid the checklist entirely.

- [ ] Self-hosted (`docker-compose up`) writes and reads traces end to end, not just SaaS

santthosh:
Closes #471. Stage 3 of #467, landing after #470 built the ledger.

#470 builds the `FindingOutcome` ledger in memory, where it dies with the invocation. The dashboard (#472) and the audit export (#295) both need it durable.

## A separate table, and the reason is correctness

I verified the issue's argument before designing around it. `queryByPR` uses `begins_with(prNumberCommitSha, :pr)` and every call site passes `${prNumber}#`:

| call site | limit |
|---|---|
| `disposition-writer.ts:398` | 10 |
| `review-processor.ts:88 / :165 / :557` | 5 |
| Lambda `review-agent.ts` | 5 |

A `42#abc123#TRACE` row matches that prefix **by construction**, so it would be returned as if it were a review — and under those small limits it could evict a real review from the lookup feeding `previousFindings`, the delta and FP-B. That is a silent correctness regression inside the review pipeline, not a query annoyance. The separate table also keeps traces clear of the 400KB item cap.

## Retention and size

| | SaaS | Self-hosted |
|---|---|---|
| store | `mergewatch-review-traces-{stage}` | `review_traces` |
| retention | 30-day TTL, **enabled** | `expires_at`; no native TTL — operators prune |
| cap | 500 outcomes, truncation marked | same |

30 days is deliberately shorter than a review's: a trace is a debugging and trust artifact, not a system of record (#295 is that).

Worth flagging: **the reviews table's `TimeToLiveSpecification` is commented out**, so its documented 90-day TTL is not actually active. I enabled TTL properly on the new table rather than copying that.

**When truncating, non-surfaced outcomes are kept first.** Surfaced findings are already visible in the PR comment, so the dropped, merged and demoted ones are the only thing the trace adds — keeping those is what makes a truncated trace still worth reading. A #401-shaped runaway (~1,430 near-empty findings from one response) must not write an unbounded record, and the remainder must never read as complete.

## Two findings that shrank the ticket

- **Self-hosted needs no new env var.** Postgres gets `review_traces` from the Drizzle schema, so there is no table-name indirection to forget. `docker-compose.yml` and `.env.example` need nothing, and the issue's deployment checklist collapses to `template.yaml` alone.
- **The Amplify SSR grant already covers it.** That inline policy is on `table/mergewatch-*` (recorded at `template.yaml:1270`), so the new table is in scope with no IAM change. `next.config.js` is untouched because nothing in the dashboard reads traces yet — that belongs with #472, which will need it.

`REVIEW_TRACES_TABLE` is added to the **#181 Globals regression guard** in `template.test.ts`: a trace writer resolving to the unsuffixed default would `ResourceNotFoundException` into a void, which is exactly the failure #181 was about.

## Safety

Writes are best-effort at both runtimes — after the comment is posted, wrapped in a catch. The developer getting their review matters; the debugging artifact does not. On self-hosted the store is optional, so a deployment that has not wired it is simply unaffected.

## Verification

```
build             12/12
typecheck         20/20
test              21/21   (1,153 tests — 16 new)
migrations:check  Everything's fine
```

The migration was hand-edited to `IF NOT EXISTS` per CLAUDE.md, matching the `0016` precedent. Includes a regression test asserting `queryByPR` queries only the reviews table and so cannot reach traces regardless of how their sort key is shaped.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -117,6 +117,32 @@ Implementations:
 | `storage-dynamo` | `DynamoReviewStore` | DynamoDB. PK=`repoFullName`, SK=`prNumberCommitSha`. 90-day TTL. |
 | `storage-postgres` | `PostgresReviewStore` | Postgres. Composite PK `(repo_full_name, pr_number_commit_sha)`. |
 
+### IReviewTraceStore
+
+Persists the per-review filter outcome ledger (#470/#471) — why each finding was
+dropped, merged, or demoted. Read by the dashboard (#472) and a future audit
+export (#295).
+
+**It has its own table, and that is a correctness requirement rather than a
+preference.** `queryByPR` matches `begins_with(prNumberCommitSha, "42#")`, so a
+trace keyed `42#abc123#TRACE` beside the review would come back as if it were a
+review — and every call site passes a small `Limit`, so trace rows could evict
+real reviews from the previous-review lookup that feeds `previousFindings`, the
+delta and FP-B.
+
+| | SaaS | Self-hosted |
+|---|---|---|
+| store | `mergewatch-review-traces-{stage}` | `review_traces` |
+| retention | 30-day Dynamo TTL, **enabled** | `expires_at` column; **no** native TTL — operators prune |
+| size guard | 500 outcomes per review, truncation explicitly marked | same |
+
+Retention is deliberately shorter than a review's 90 days: a trace is a
+debugging and trust artifact, not a system of record. Writes are best-effort at
+both runtimes — a trace must never be able to fail a review.
+
+Self-hosted needs **no new environment variable**: the table comes from the
+Drizzle schema, so there is no table-name indirection to forget.
+
 ### IGitHubAuthProvider
 
 ```typescript
--- a/infra/template.yaml
+++ b/infra/template.yaml
@@ -104,6 +104,11 @@ Globals:
         # DynamoDB table names — injected so Lambda code doesn't hardcode them
         INSTALLATIONS_TABLE: !Ref InstallationsTable
         REVIEWS_TABLE: !Ref ReviewsTable
+        # #471 — the filter outcome ledger. A SEPARATE table, not a suffixed
+        # sort key on ReviewsTable: queryByPR matches begins_with("42#"), so a
+        # trace row there would come back as a review and could evict a real
+        # one from the previous-review lookup under its small Limit.
+        REVIEW_TRACES_TABLE: !Ref ReviewTracesTable
         API_KEYS_TABLE: !Ref ApiKeysTable
         SESSIONS_TABLE: !Ref SessionsTable
         # FB-A / FB-E: shared across ReviewAgent (writes) and InsightsRollup (reads/writes).
@@ -431,6 +436,7 @@ Resources:
           STAGE: !Ref Stage
           INSTALLATIONS_TABLE: !Ref InstallationsTable
           REVIEWS_TABLE: !Ref ReviewsTable
+          REVIEW_TRACES_TABLE: !Ref ReviewTracesTable
           API_KEYS_TABLE: !Ref ApiKeysTable
           SESSIONS_TABLE: !Ref SessionsTable
           DEFAULT_BEDROCK_MODEL_ID: !Ref DefaultBedrockModelId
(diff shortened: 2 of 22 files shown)


santthosh:
Both warnings taken — and the convention claim in them is correct, which is what settled it.

`finding-disposition-store.ts` and `review-store.ts` already guard jsonb and Item reads with `Array.isArray` in **both** backends:

```
storage-postgres/finding-disposition-store.ts:200  Array.isArray(r.sigTokens) ? …
storage-postgres/review-store.ts:153               Array.isArray(row.inlineResolvedKeys) …
storage-dynamo/finding-disposition-store.ts:241    if (Array.isArray(it.sigTokens)) …
```

The new trace stores were the inconsistent ones, not the reviewer.

**The consequence is real rather than theoretical.** `outcomes` is iterated by the dashboard (#472) and the audit export (#295). A jsonb column accepts any shape regardless of its declared type, so a partial write or an older item shape would pass the cast and crash the *consumer* rather than failing where the data is read.

Fixed in both, in the direction that suits a best-effort artifact — a missing trace is a nuisance, a crashing dashboard is not:

| backend | malformed `outcomes` |
|---|---|
| Postgres | degrades to an empty ledger, warns |
| Dynamo | treated as absent (`null`), warns |

Also type-checked `truncated` and `totalOutcomes` rather than passing them through, so a stray string cannot reach a consumer expecting a number — the warning called that out and it was a fair catch.

4 new tests: non-array outcomes, absent outcomes, and wrong-typed truncation fields. `typecheck 20/20 · test 21/21`, now 20 new tests on this PR.

PR #481, opened 2026-08-26
feat(dashboard): render findings, summary and merge score on review detail

santthosh:
Part A of #472. The issue notes this part depends on nothing, so it lands on its own rather than behind the trace plumbing — it has been a live gap on every review detail page since the feature shipped.

## The bug is prop plumbing, not missing data

`ReviewDetail.tsx` rendered no findings, no summary and no merge score. The store and `/api/reviews/[id]` both return all three; `page.tsx` fetched the review and passed a subset. The last line of the component was, verbatim:

```tsx
{/* Future: reasoning flow, agent traces, etc. */}
```

## Decisions

- **Severity styling is borrowed from `ReviewDrawer`**, not re-invented, so the list view and the detail view describe the same review the same way.
- **Ordering is stable within a severity.** The orchestrator already ranked the findings; re-sorting inside a severity would discard that ranking for nothing.
- **Logic lives in `lib/review-findings.ts`.** The dashboard suite tests pure functions in `lib/` rather than rendering components, so that is where the testable part goes.
- **An unverified critical gets an `unverified` chip**, mirroring FP-L's split in the PR comment — a demoted critical is advisory and should not read as blocking here either.
- **A completed review with no findings says so**, and points at the decision trail (Part C) rather than rendering nothing.

## One thing worth recording

Importing `mergeScoreMeta` from `@mergewatch/core` into a **client** component breaks the build:

```
UnhandledSchemeError: Reading from "node:fs/promises" is not handled
../core/dist/context/safe-path.js -> ../core/dist/index.js -> ./components/ReviewDetail.tsx
```

Core's index pulls `context/safe-path`, which is server-only. So the verdict wording is duplicated in the dashboard lib — and **kept honest by a test** that imports the real `mergeScoreMeta` (in Node, where core resolves fine) and asserts both agree for all five scores. Drift fails CI rather than shipping a dashboard whose verdict contradicts the PR comment.

That test also caught its own purpose: it verified the hand-transcribed emoji were right.

## Verification

```
build      12/12
typecheck  20/20
test       21/21   (1,167 tests — 14 new)
```

Parts B and C follow in a second PR, now that #471 is in production.


santthosh:
Real bug, and tracing it found the divergence was the smaller half of the problem.

`MERGE_SCORE_META` is keyed 1–5. Core clamped but never rounded, so a fractional score **missed the table entirely**:

```
$ node -e "mergeScoreMeta(2.5)"
{"score":2.5}        // no emoji, no label
```

which `renderMergeScore` turns into `undefined **2.5/5 — undefined**` in the rendered PR comment. And it is reachable: `runOrchestratorAgent` does `Math.max(1, Math.min(5, parsed.mergeScore ?? 3))` with **no rounding**, so a model returning `2.5` lands there.

**So I took the second of the two suggestions, not the first.** Removing `Math.round` from the dashboard would have matched core exactly — by replicating core's bug on a second surface. Rounding belongs in core, where it fixes the comment path too.

The other half of the finding was the sharper observation: **the drift guard only checked integers, which is precisely why it could not see this.** A guard that can only fail on inputs the two implementations already agree on is not a guard. It now covers fractional and out-of-range values:

```ts
for (const score of [1, 1.4, 2, 2.5, 3, 3.7, 4, 4.9, 5, 0, 99]) {
  expect(mergeScoreMeta(score)).toEqual(core.mergeScoreMeta(score));
}
```

Core also gains a test asserting no score can produce a label-less verdict, and that a fractional score renders as a whole number in the comment rather than `undefined`.

Worth noting this bug predates the PR — it was latent in `renderMergeScore` and my dashboard copy accidentally masked it on one surface while exposing the difference on the other.

`build 12/12 · typecheck 20/20 · test 21/21`, 17 new tests.

Issue #477, opened 2026-08-25
Evidence follow-up: convergence is destroyed by the orchestrator, and cited code can break the comment


santthosh:
Follow-up to #469 (shipped in #476). Two defects found auditing the merged PR. The first means one of the three evidence elements never reaches production.

## 1. Cross-agent convergence is destroyed by the orchestrator

`evidence.agents` is set by `dedupeCrossAgentByLine` (`finding-clustering.ts:349`) — FP-C, which runs **before** `runOrchestratorAgent`. The orchestrator returns findings reconstructed entirely from the LLM's JSON response:

```ts
return {
  findings: parsed.findings ?? [],     // reviewer.ts:1418
  mergeScore: ...,
  mergeScoreReason: ...,
};
```

`ORCHESTRATOR_SCHEMA` (`agents/schemas.ts:49`) declares no `evidence` property, and nothing downstream re-attaches it — `grep -n evidence reviewer.ts` shows exactly two write sites after this point, `reason` at `:2320` and `code` at `:2369`.

**Result:** `security + bugs agreed independently` effectively never renders. The pipeline order is:

```
agents → FP-C (sets evidence.agents) → orchestrator (DESTROYS it) → W10 → filters
       → grounding → verifyFindings (sets reason) → withEvidenceCode (sets code)
```

Worse than a clean absence: `ORCHESTRATOR_SCHEMA` is deliberately permissive (no `additionalProperties: false`), so the model *may* sometimes echo the field back. That makes the behaviour intermittent and effectively undiagnosable in the field, rather than reliably missing.

This also removes the element the #469 rationale called "the strongest confidence signal in the system" — two agents independently reaching the same line — which was the whole reason FP-C's merge was worth instrumenting.

### Why the tests pass anyway

Three suites cover this feature and none crosses the seam where it breaks:

| Test | What it actually proves |
|---|---|
| `finding-clustering.test.ts` "records which agents converged" | `dedupeCrossAgentByLine` sets the field, in isolation |
| `comment-formatter.test.ts` "renders all three elements" | the renderer displays a hand-constructed `evidence.agents` |
| `reviewer.test.ts:3839` "preserves evidence already attached upstream" | `withEvidenceCode` merges rather than overwrites a hand-injected value |

Each tests one side of the break. E2E-101's convergence checkbox is marked `unit-test-gated`, so it does not catch it either.

[...]

PR #482, opened 2026-08-26
feat(dashboard): per-finding evidence panel and the filtered-findings trail

### LINKED ISSUE #472: Dashboard: render findings, per-finding evidence, and the filtered-findings trail

santthosh:
Closes #472. Parts B and C, now that #471's ledger is in production. Part A shipped in #481.

## Part B — the evidence panel

Everything #469 puts in the PR comment, uncollapsed, **plus the two elements deliberately cut from that surface** as too jargon-heavy for it:

- **Confidence vs the floor in force** — `82% (floor 75)`. "82%" alone does not say whether it nearly missed the cut.
- **Grounding result** — unactionable in a comment, but it is exactly the answer someone gets when they deliberately open the evidence.

Uncollapsed on purpose. A reader who opened a review detail page to find out *why* a finding exists should not have to click again to see the proof — that is the same mistake as hiding it in the comment.

## Part C — the filtered-findings trail

Every non-surfaced outcome, grouped by stage, **judgement-bearing stages first**:

| first | then |
|---|---|
| `finding-verify`, `orchestrator`, `grounding` | `confidence-floor`, `min-severity`, `line-proximity`, … |

Someone auditing a missing finding cares about where a *model made a call* — that is where a defensible finding can be wrongly discarded. A confidence floor doing exactly what it is configured to do is far less interesting, and putting it first buries the part worth reading.

## Decisions worth calling out

- **`getReviewTrace` is required on `IDashboardReviewStore`, not optional.** The dashboard is part of the self-hosted stack, so a Dynamo-only read would leave those users with an empty trail *and no error* — and an empty trail reads as "nothing was filtered", which is worse than the feature not existing. Making it required turned shipping one backend into a compile error, and both were implemented against that error:

```
error TS2420: Class 'PostgresDashboardReviewStore' incorrectly implements
  interface 'IDashboardReviewStore'. Property 'getReviewTrace' is missing.
```

- **Demotions are not removals.** A demoted critical still renders above as advisory, so counting it in the funnel would make the funnel disagree with the findings list. It gets its own chip colour and its own line.
- **Orchestrator drops render "not explained"**, never a blank cell. A blank reads as "nothing to say" — the exact coverage illusion this effort exists to remove.
- **A failed fetch is distinguished from an absent trace.** A network failure is not evidence that nothing was filtered, and must not be presented as if it were.
- **Surfaced findings are excluded from the trail** — they are the result, rendered above; repeating them would read as a second, competing list.

## Client-bundle constraint

The dashboard declares its own `FilterStage` union rather than importing core — core's index pulls `context/safe-path` (`node:fs/promises`) and a client-component import fails the webpack build, as #481 found. A **drift test imports core in Node** and asserts the copies still agree, so divergence fails CI.

## Verification

```
build      12/12
typecheck  20/20
test       21/21   (1,189 tests — 22 new)
```

Trace fetching is client-side via `/api/reviews/[id]/trace`, per the CLAUDE.md Amplify SSR caveat — and a separate route from `/api/reviews/[id]` so the review renders without waiting on a ledger that can run to hundreds of rows. `DYNAMODB_TABLE_REVIEW_TRACES` is declared in `next.config.js`, which is the #315 trap the issue points at.


santthosh:
Both concerns land, though the second is not the bug it was reported as.

## ✅ Stageless outcomes — real, and it contradicted my own comment

The line directly above said *"surface it rather than silently dropping the row"*, and then the code misattributed the row to the orchestrator. That is a different lie from dropping it, and no better.

It matters more than a labelling nit because **#470's recorder emits `stage: undefined` on purpose** — an entry that reached no terminal stage surfaces as a wiring gap rather than being quietly balanced away. Bucketing those into a real stage destroys exactly the signal that was built for.

Fixed with a `stage-not-recorded` sentinel: sorted last (a gap is not a decision), labelled, **not** marked judgement-bearing, and **not** borrowing the orchestrator's "#473 does not report why" explanation — which would be a false attribution of its own.

## ❌ `remaining` cannot go negative — ✅ but checking it found a real error next door

The arithmetic rules it out. Every outcome lands in exactly one stage group, and removals exclude both surfaced and demoted, so removals can never exceed `raw − surfaced − demoted`.

**But verifying that surfaced a genuine error in my own docstring.** It claimed:

> the last step's `remaining` equals the number of surfaced findings

That is only true for reviews with **no demotions**. A demotion is not a removal, so the real value is `surfaced + demoted`. Wrong comment, right code — and my original test only covered the no-demotion case, so it could never have caught it.

Corrected, with three invariants now pinned instead of asserted in prose:

```
never negative · ends at surfaced + demoted · removals <= raw - surfaced - demoted
```

## The drift test earned its keep

Adding the sentinel broke it — correctly. It now excludes the sentinel from the core comparison (it is deliberately not a pipeline stage) while asserting it exists in the label map, so a real stage added to core still fails CI.

`build 12/12 · typecheck 20/20 · test 21/21` — 31 new tests on this PR.

santthosh:
One holds, one does not.

## ✅ Per-element validation — real, with a concrete crash

`Array.isArray` checks the container and stops there, but `outcomes` is `jsonb` in Postgres and an unvalidated map in DynamoDB, so the **elements** are equally unconstrained. The renderer calls `e.agents.join(…)`, and a row missing that field throws:

```
Cannot read properties of undefined (reading 'length')
```

`isUsableOutcome` now requires exactly the fields the renderer dereferences — nothing stricter, which would reject rows that display perfectly well. Both dashboard reads filter through it.

**Malformed rows are dropped rather than failing the whole read** — losing one corrupt row should not cost the reader the other ninety-nine. But a partial trail must never read as complete, so a read that loses rows marks the trace `truncated` with the true total, reusing the affordance the UI already has for the size cap. Corruption and truncation tell the reader the same thing: you are not seeing everything.

## ❌ `merged` in the funnel — the fix would cause the bug it describes

When A and B merge, **B is recorded `merged` and A survives to surface**. So:

| | |
|---|---|
| `raw` | 2 |
| visible findings | 1 |
| current: removed=1 → `remaining` | **1** ✓ |
| suggested: removed=0 → `remaining` | **2** ✗ |

Excluding merges would report 2 remaining against 1 visible finding — exactly the disagreement the finding says it is preventing. A merged finding *is* gone as a separate row; that is what merging means.

The identity `remaining === surfaced + demoted` holds with merges, drops and demotions together, and there is now a test using all three at once:

```ts
expect(f.steps.at(-1)!.remaining).toBe(f.surfaced + demotedCount(outcomes));
```

Worth noting this is the **third round on this same line**, each with a different mechanism — first "can go negative", then "can fall below surfaced + demoted", now "double-counts merges". The arithmetic has not changed, and it is now pinned by tests rather than argued each time.

`build 12/12 · typecheck 20/20 · test 21/21` — 40 new tests on this PR.

santthosh:
Critical holds and is fixed. The warning does not. The info note is fair but not worth acting on.

## ✅ Unhandled store rejection — real

`canAccessRepo` was wrapped; the `getReviewTrace` call below it was not. A store error propagated as an unformatted 500 with nothing logged.

The finding's own analysis of the client side is correct and worth repeating: `FilterTrail` already distinguishes a failed fetch from an absent trace, so the **UI never claims "nothing was filtered" on an infrastructure fault**. The gap was purely server-side observability — and that is the same shape as the #471 bug I found earlier today, where a silent failure let a broken feature look healthy in production for hours. Now logged, with an explicit 500.

Also worth noting: **`/api/reviews/[id]` has the same unwrapped store call.** I have not changed it — it is untouched by this PR and fixing it would be a drive-by — but it is the same gap and deserves its own change.

(Incidentally, this finding carries `error-handling + security agreed independently` — the cross-agent convergence from #477, working on a real review.)

## ❌ Truncation banner — the numbers are already exact

The banner reads **"Showing {outcomes.length} of {totalOutcomes}"**, where the left side is what the reader can actually see and the right is what the review produced:

| case | shown | of | true? |
|---|---|---|---|
| 3 of 100 corrupt | 97 | 100 | ✅ exactly right |
| capped at 500 of 1430, 3 corrupt | 497 | 1430 | ✅ exactly right |

"Showing 97 of 100" after three corrupt rows is precisely true — 97 is what is displayed. The suggested `retainedOutcomes` field would add a number the reader has no use for. I have added a comment stating the semantics so it does not have to be re-derived next time.

## 🔵 Unrecognised stages — acknowledged, not acted on

Correct that an outcome carrying a stage the deployed dashboard does not know renders under "Stage not recorded" until redeploy. The finding also correctly scopes it as a deployment-window concern rather than a persistent bug, and the drift test catches divergence in CI.

Declining the `console.warn` specifically: `filter-trail.ts` is used from a client component, so that warning would land in the **viewer's browser console**, which no operator reads. It would create the appearance of observability without any.

`build 12/12 · typecheck 20/20 · test 21/21`.

PR #487, opened 2026-08-26
feat(dashboard): drawer carries the full review; one URL shape for both runtimes

### LINKED ISSUE #486: Drawer should carry the full review: evidence and decision trail, one implementation
Follow-up to #472. The work landed on a page users cannot reach.

## The problem

Clicking a review in the dashboard opens **`ReviewDrawer`**. That is the only interaction path — `grep` finds **zero** links to `/dashboard/reviews/<id>` anywhere in the components. The drawer's only outbound links go to GitHub (commit, comment, PR).

Meanwhile #472 put the per-finding evidence panel and the decision trail on `ReviewDetail`, which is reachable only by hand-assembling a URL or clicking through from a GitHub comment.

So the surface people use has **less** information than the one they do not.

## Why it happened

#472 framed every part around `ReviewDetail.tsx` and cast the drawer as the reference to borrow *from*:

> `ReviewDetail.tsx` renders no findings, no summary, and no merge score at all… `ReviewDrawer.tsx` already renders findings with severity dots and confidence badges and can be borrowed from.

That framing was followed without asking how a user reaches `ReviewDetail.tsx`.

## What to do

**The drawer carries everything.** Not by duplicating markup — by extracting what is currently locked inside `ReviewDetail.tsx` into shared components:

```
components/FindingEvidence.tsx   <- extract EvidencePanel, FindingsSection, SeverityDot
components/FilterTrail.tsx       <- already standalone

ReviewDrawer  -> imports both    (where people click)
ReviewDetail  -> imports both    (thin wrapper; see below)
```

One implementation, two mount points. This also **removes an existing fork**: `ReviewDetail`'s severity styling was copied from the drawer during #472, so sharing collapses two copies into one.

## The route must keep resolving

`/dashboard/reviews/[id]` cannot simply be deleted. It is the link target for three surfaces already live:

| surface | reference |
|---|---|
| every review comment footer | `[View full details](reviewDetailUrl)` — `comment-formatter.ts:855` |
| Check Run `details_url` | `review-agent.ts:1218` — the "Details" link in the PR merge box |
| #468 truncation notice | `[View the full review](reviewDetailUrl)` — the only recovery path when a comment is cut |

Deleting the route 404s every comment MergeWatch has already posted. So it stays, rendering the **same shared components** — a thin wrapper, not a second implementation.

## Known constraint

Width. The trail's stage groups and cited-code blocks are laid out for a page; in a narrow panel they need to wrap or collapse. A styling problem, not a reason to withhold the content.

## Acceptance criteria

- [ ] Evidence panel renders per finding **in the drawer**
- [ ] Decision trail renders **in the drawer**
- [ ] Exactly one implementation of each — no copied markup between drawer and page
- [ ] The drawer's pre-existing findings markup is replaced by the shared component, not left alongside it
- [ ] `/dashboard/reviews/[id]` still resolves and renders the same content
- [ ] Code blocks and stage groups remain readable at drawer width

santthosh:
Closes #486.

## The gap

#472's evidence panel and decision trail landed on `ReviewDetail` — a page **nothing in the dashboard links to**. `grep` finds zero references to `/dashboard/reviews/<id>` in any component: clicking a review calls `onSelect()` and opens the drawer, full stop.

So the surface people use had *less* information than the one they could not reach. That happened because #472 framed every part around `ReviewDetail` and cast the drawer as the thing to borrow *from*. I followed the framing without asking how a user reaches `ReviewDetail`, which was the question that mattered.

## Extracted, not duplicated

`SeverityDot`, `EvidencePanel` and `FindingsSection` were **local functions inside `ReviewDetail`**. Nothing about them is page-specific — that accident is the only reason the drawer could not show evidence.

```
components/FindingEvidence.tsx   ← the three, now shared
components/FilterTrail.tsx       ← already standalone

ReviewDrawer  → mounts both   (where people click)
ReviewDetail  → mounts both   (link target; see below)
```

**This collapses a fork rather than creating one.** `ReviewDetail`'s severity palette was copied from the drawer during #472; the drawer's now-dead copy is deleted here. One palette, one findings renderer, one evidence panel — which is why the diff is net **−194 lines** across the two components.

## The route stays, and has to

It is the target of three live surfaces:

| surface | reference |
|---|---|
| every review comment footer | `comment-formatter.ts:855` |
| Check Run `details_url` | `review-agent.ts:1218` |
| #468 truncation notice | the only recovery path when a comment is cut |

Deleting it would 404 every comment MergeWatch has already posted. It is now a thin mount of the same shared components — not a second implementation.

## Self-hosted bug found while answering "does this hardcode endpoints?"

It does not — the trail fetches a **relative** path, so it adapts to any host. But the two runtimes built **different** detail URLs:

```
SaaS:        /dashboard/reviews/owner%2Frepo%3A42%23abc     1 segment  ✅
self-hosted: /dashboard/reviews/owner%2Frepo/42#abc         2 segments ❌
```

The route is a single `[id]` segment, so the self-hosted form **404s** — and the unencoded `#` makes the commit SHA a browser fragment:

```
path     : /dashboard/reviews/owner%2Frepo/42
fragment : abc123   ← never sent to the server
```

So *View full details* has **never worked on self-hosted**. That also means **#468's truncation-recovery link was dead there** — the one path to content cut from an oversized comment.

Both runtimes now call `buildReviewDetailUrl` in core, with a test that round-trips through the route's **own parser** so the shapes cannot drift again.

## Verification

```
build      12/12
typecheck  20/20
test       21/21
```

Not claimed: a live `docker-compose` click-through. The URL fix is covered by the round-trip test, and the trail's relative fetch is host-agnostic by construction, but I have not stood a compose stack up.

### DIFF
--- a/packages/core/src/comment-formatter.test.ts
+++ b/packages/core/src/comment-formatter.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from 'vitest';
-import { formatReviewComment, buildWorkDoneSection, countBlockingCriticals, buildCheckTitle, escapeUserContent, COMMENT_BODY_BUDGET, mergeScoreMeta, type Finding } from './comment-formatter.js';
+import { formatReviewComment, buildWorkDoneSection, countBlockingCriticals, buildCheckTitle, escapeUserContent, COMMENT_BODY_BUDGET, mergeScoreMeta, buildReviewDetailUrl, type Finding } from './comment-formatter.js';
 
 // ─── Helpers ──────────────────────────────────────────────────────────────────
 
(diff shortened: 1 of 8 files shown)


santthosh:
Taken — and the alternative in the suggestion was the better half of it.

**`ReviewDetail.id` already IS the canonical `prNumberCommitSha`** — the same key the trace is stored under. So rebuilding it from `prNumber` + `commitSha` was unnecessary as well as lossy, and I should not have reconstructed something that was already sitting on the object.

**The failure mode is real and the type does not protect against it.** `commitSha` is derived as `prNumberCommitSha.split("#")[1] ?? ""` in both the list API (`route.ts:78`) and `ReviewsClient` (`|| ""`), so a stored key without a `#` produces `repo:42#`. That lookup finds nothing, and `FilterTrail` then renders:

> No decision trail was recorded for this review.

Which would be **a false statement, not an absent trail** — a lookup failure presented as an absence. That is precisely the failure class #467 exists to remove, so it matters more here than the usual undefined-interpolation concern.

Both surfaces now key on the canonical id and guard the fields they interpolate, so a partially-hydrated review renders **no trail section at all** rather than a confident denial.

`build 12/12 · typecheck 20/20 · test 21/21`

PR #499, opened 2026-08-26
fix(dashboard): keep the selected org in a user-scoped cookie

### LINKED ISSUE #498: Opening a review detail page silently resets the selected organization to the default
## Symptom

Select an organization in the sidebar (e.g. `mergewatch`), open a review's detail page, and the org switcher silently reverts to the personal/default org (`santthosh`). Nothing indicates a change happened.

## Mechanism

Org selection is **URL state**, not stored state. `DashboardShell.tsx:32`:

```ts
const orgParam = searchParams.get("org");
const activeInstallationId = orgParam
  ? Number(orgParam)
  : installations[0]?.id ?? 0;      // ← no param means "first installation"
```

For most users `installations[0]` is their personal account, so **any navigation that drops `?org=` silently resets the selection to personal.**

A helper that does the right thing already exists — `Sidenav.tsx:108`:

```ts
/** Append ?org= to a path when a non-default installation is selected */
function navHref(base: string) {
  return orgParam ? `${base}?org=${orgParam}` : base;
}
```

…but it is applied in exactly **one** place: the sidenav's main nav items (`Sidenav.tsx:326`). Every other in-app link bypasses it:

| File:line | Link |
|---|---|
| `components/ReviewDrawer.tsx:439` | → review detail — **the reported path** |
| `components/ReviewDetail.tsx:204` | "Back to dashboard" — so returning also lands on the wrong org |
| `components/DashboardContent.tsx:84` | → repositories |
| `components/Header.tsx:90` | → profile |
| `components/layout/Sidenav.tsx:137, 165, 168, 172` | logo → `/dashboard` (4 responsive variants) |

## Why this is worse than a wrong label

The review itself renders correctly — it is fetched by `repoFullName` and gated by `canAccessRepo`, independent of the switcher. So the page looks right while the sidebar quietly says something else.

The damage lands on the *next* action. The user believes they are in `mergewatch`, clicks "Back to dashboard", and gets the personal org's reviews — reading one org's data under another org's label. Nothing errors, and nothing announces the switch.

## Fix — persist the selection, do not thread it through links

Patching the eight call sites does not hold: the ninth link added next month forgets it again, which is how this one arrived. Persist the selection instead, so navigation cannot lose it and no per-link discipline is required. The URL param stays as a deep-link override.

### Use a cookie, not localStorage/sessionStorage

The dashboard is server-component heavy (`export const dynamic = "force-dynamic"` throughout), so a client-only store produces a **hydration flash**: the server renders the default org, the client reads storage and corrects it, and the user watches the switcher flip on every page load. That is a more visible version of the bug being fixed.

A cookie is readable server-side, so the correct org renders on first paint.

Trade-off worth accepting knowingly: a cookie is shared across tabs, where `sessionStorage` would allow a different org per tab. Cross-tab consistency is the better default here; revisit if anyone actually wants per-tab orgs.

### Precedence

```
?org= param  >  cookie  >  explicit fallback
```

A URL param both wins and refreshes the cookie, so shared deep links keep working and switch the session to that org — which is what a person clicking a shared link expects.

### Clearing on sign-out and sign-in

Clear the cookie in the NextAuth `signOut` flow, as requested. But **do not rely on clearing as the correctness boundary** — store the owning user alongside the selection:

```ts
{ userId: <githubUserId>, installationId: <id> }
```

and ignore the cookie whenever `userId !== session.githubUserId`. Then a missed clear, a shared browser, or a stale cookie from a previous account is inert by construction rather than by hoping the clear ran. Clearing becomes hygiene; the mismatch check is the guarantee.

### Validate before trusting

A persisted id must be checked against the current user's installations on every read — access can be revoked between sessions, so a stored id can become one the user cannot see. This matters more with persistence than it did with a URL param, because the value now survives indefinitely.

And the existing fallback is part of the bug:

```ts
: installations[0]?.id ?? 0    // DashboardShell.tsx:32
```

Silently picking "first in the list" is what turns a missing or invalid selection into a *wrong answer* instead of a visible one. Whatever the fallback becomes, the switcher should not assert an org the user did not choose without that being apparent.

## Acceptance criteria

- [ ] Opening a review detail page from the drawer preserves the selected org
- [ ] "Back to dashboard" from a review returns to the same org, not the default
- [ ] The selection survives navigation to any dashboard route with no `?org=` in the URL
- [ ] No hydration flash — the correct org renders on first paint, server-side
- [ ] A `?org=` deep link wins over the cookie and updates it
- [ ] Signing out clears the cookie
- [ ] A cookie whose `userId` does not match the session is ignored, tested with a second account on the same browser
- [ ] A stored installationId the user can no longer access does not silently resolve to a different org
- [ ] Deep-linking to a review with no `?org=` still renders the review correctly
- [ ] Switching org from a review detail page stays on that page (existing `handleSwitch` behaviour, `DashboardShell.tsx:42`)

santthosh:
Closes #498

## The bug

Selecting an org (e.g. `mergewatch`) and opening a review's detail page silently reverted the switcher to the personal account.

Org selection lived only in the URL:

```ts
const activeInstallationId = orgParam ? Number(orgParam) : installations[0]?.id ?? 0;
// DashboardShell.tsx:32
```

Any link omitting `?org=` reset it, and the fallback to `installations[0]` turned the omission into a *wrong answer* rather than a visible one. A helper that appends the param already existed (`Sidenav.tsx:108`) — wired to exactly one place, the sidenav's nav items. Eight other links bypassed it, including the review link and "Back to dashboard".

The review itself renders correctly (fetched by `repoFullName`, gated by `canAccessRepo`), so the page looked right while the sidebar said something else. The damage landed on the next click: "Back to dashboard" then showed the personal org's reviews under the belief you were in `mergewatch`.

## The fix

Threading the param through those eight links would not hold — the ninth link added next month forgets it again, which is how this one arrived. The selection now persists in a cookie, so navigation cannot lose it.

**A cookie, not localStorage.** The dashboard is server-rendered throughout (`force-dynamic`). A client-only store renders the default on the server and corrects it on hydration, flipping the switcher on every page load — a more visible version of the bug being fixed. The layout reads the cookie server-side, so first paint is already correct.

**Precedence:** `?org=` → cookie → first installation. A shared deep link still wins and refreshes the cookie, so it switches the reader to the org it names.

**User-scoped.** The cookie carries the GitHub user id that wrote it and is ignored when it does not match the session. Sign-out clears it at all four `signOut()` call sites — but that is hygiene, not the boundary: a missed clear, a shared browser, or a cookie from a previous account is inert by construction.

**Validated, not trusted.** Both the param and the cookie are checked against the installations the account currently has. Access can be revoked between visits, and resolving a stale id would show one org's label over another org's data. A cookie that cannot resolve is dropped rather than re-read on every navigation forever.

## Tests

19 new tests in `lib/org-selection.test.ts`. All resolution logic is pure and lives in `lib/`, which is where this package's vitest config can reach (`include: ["lib/**/*.test.ts"]`, node env) — components stay thin wiring.

Covered: precedence ordering, user mismatch (shared browser), revoked installation, malformed and undecodable cookie values, colon-in-user-id, missing session user, empty installations, and stale-param-with-valid-cookie.

One of these caught a real bug during development: a present-but-unparseable cookie was not marked stale, so a corrupt value would have been re-read forever. Fixed in the same commit.

## Verification

- `pnpm run build` — 12/12 green
- `pnpm run test` — 21/21 tasks; dashboard suite 152 tests across 9 files
- `npx tsc --noEmit` in `packages/dashboard` — clean (note: the dashboard has no `typecheck` script, so `pnpm run typecheck` skips it — ran directly)

## Not included

`packages/dashboard/tsconfig.tsbuildinfo` is tracked in git and churns on every build. Left out of this diff deliberately; worth un-tracking separately.

### DIFF
--- a/packages/dashboard/app/dashboard/layout.tsx
+++ b/packages/dashboard/app/dashboard/layout.tsx
@@ -1,9 +1,11 @@
 import { redirect } from "next/navigation";
+import { cookies } from "next/headers";
 import { getServerSession } from "next-auth";
 import { authOptions } from "@/lib/auth";
 import { fetchUserInstallations, TokenExpiredError } from "@/lib/github-repos";
 import DashboardShell from "@/components/layout/DashboardShell";
 import type { InstallationInfo } from "@/components/layout/DashboardShell";
+import { ORG_COOKIE } from "@/lib/org-selection";
 
 export default async function DashboardLayout({
   children,
@@ -41,11 +43,18 @@ export default async function DashboardLayout({
     type: i.account.type,
   }));
 
+  // #498 — read the selection server-side so the correct org renders on first
+  // paint. Resolving it in the client would render the default on the server
+  // and correct it on hydration, flipping the switcher on every page load.
+  const orgCookie = (await cookies()).get(ORG_COOKIE)?.value;
+
   return (
     <DashboardShell
       userName={session.user?.name ?? session.user?.email ?? ""}
       userImage={session.user?.image}
       installations={installationInfos}
+      orgCookie={orgCookie}
+      githubUserId={(session as any).githubUserId as string | undefined}
     >
       {children}
     </DashboardShell>
--- a/packages/dashboard/app/signout/page.tsx
+++ b/packages/dashboard/app/signout/page.tsx
@@ -1,10 +1,12 @@
 "use client";
 
 import { signOut } from "next-auth/react";
+import { clearOrgCookie } from "@/lib/org-selection";
 import { useEffect } from "react";
 
 export default function SignOutPage() {
   useEffect(() => {
+    clearOrgCookie();
     signOut({ callbackUrl: "/" });
   }, []);
 
--- a/packages/dashboard/components/Header.tsx
+++ b/packages/dashboard/components/Header.tsx
@@ -3,6 +3,7 @@
 import { useState, useRef, useEffect } from "react";
 import Link from "next/link";
 import { signOut } from "next-auth/react";
+import { clearOrgCookie } from "@/lib/org-selection";
 import { Menu, User } from "lucide-react";
 
 interface HeaderProps {
@@ -98,7 +99,7 @@ export default function Header({ userName, userImage, onMenuToggle }: HeaderProp
                 </div>
               </Link>
               <button
-                onClick={() => signOut({ callbackUrl: "/" })}
+                onClick={() => { clearOrgCookie(); signOut({ callbackUrl: "/" }); }}
                 className="flex w-full items-center gap-2 px-4 py-2.5 text-left text-sm text-primer-muted transition hover:bg-surface-card hover:text-red-400"
               >
                 <svg
--- a/packages/dashboard/components/layout/DashboardShell.tsx
+++ b/packages/dashboard/components/layout/DashboardShell.tsx
@@ -63,7 +63,11 @@ export default function DashboardShell({
   // switch the reader's session to the org it names.
   useEffect(() => {
     if (typeof document === "undefined") return;
-    if (resolved.staleCookie && resolved.source !== "param") {
+    // Cleared unconditionally, then re-written below when there is something
+    // to write. Guarding this on `source !== "param"` relied on the write
+    // always following, but its own guard can fail (a session with no
+    // githubUserId), and the stale cookie then survived indefinitely.
+    if (resolved.staleCookie) {
       document.cookie = serializeClearedOrgCookie();
     }
     if (githubUserId && resolved.installationId != null && resolved.source !== "cookie") {
--- a/packages/dashboard/components/layout/DashboardShell.tsx
+++ b/packages/dashboard/components/layout/DashboardShell.tsx
@@ -5,7 +5,7 @@ import { useRouter, usePathname, useSearchParams } from "next/navigation";
 import Sidenav from "./Sidenav";
 import {
   resolveOrg,
-  serializeOrgCookie,
+  writeOrgCookie,
   serializeClearedOrgCookie,
 } from "@/lib/org-selection";
 
@@ -71,7 +71,7 @@ export default function DashboardShell({
       document.cookie = serializeClearedOrgCookie();
     }
     if (githubUserId && resolved.installationId != null && resolved.source !== "cookie") {
-      document.cookie = serializeOrgCookie(githubUserId, resolved.installationId);
+      writeOrgCookie(githubUserId, resolved.installationId);
     }
   }, [githubUserId, resolved.installationId, resolved.source, resolved.staleCookie]);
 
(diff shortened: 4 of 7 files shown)


santthosh:
### Re: `resolveOrg` staleCookie when a param outranks a valid cookie

Assessed and **not applied** — the premise doesn't hold, and the suggested alternative would regress.

**The cookie is already overwritten.** The write condition in `DashboardShell` is `source !== "cookie"`. The param path returns `source: "param"`, so the write runs and replaces the cookie with the param's value. Nothing is left disagreeing with the URL.

**Flagging it stale would be a bug.** `staleCookie` means "this cannot resolve, drop it". A valid cookie that merely lost to a param is not stale, and clearing it on every deep-link visit is a different defect.

**The alternative would reintroduce #498.** Narrowing to `source === "param"` stops the write firing for `source === "default"` — so a first-time visitor's org is never persisted, and the original symptom returns for anyone who hasn't used the switcher yet.

**The one uncovered-looking path is covered.** The write is skipped when `githubUserId` is absent — and that same condition makes the cookie unusable in `resolveOrg` (`cookieUsable` requires `userId`), so `staleCookie` is true and it's cleared instead.

Pushed `1ed477a`→ latest with no behaviour change: the reasoning now lives in the effect, plus a test pinning the contract so this doesn't get "fixed" into the regression later.

For the record across three rounds: 4 findings, 2 real and fixed (stale cookie surviving the param path, `Secure` flag), 1 documented rather than applied with reasoning in-source (`HttpOnly` — it would silently break the `document.cookie` write path), 1 false positive. The `Secure` finding was right but its suggested `NODE_ENV` gate would have broken self-hosted, which runs a production build over plain HTTP — keyed on `window.location.protocol` instead.

santthosh:
### Re: cookie value not validated for length / header injection

Assessed and **not applied** — the injection vector doesn't exist, and the input isn't attacker-controlled.

**`encodeURIComponent` already neutralizes it.** The value is `encodeURIComponent(\`${userId}:${installationId}\`)`, which percent-encodes every delimiter that could break out of a cookie value. Verified:

```
input:  "9001\r\nSet-Cookie: evil=1; x"
output: "9001%0D%0ASet-Cookie%3A%20evil%3D1%3B%20x%3A222"
```

CR, LF, `;`, `=` and space all encode. There is no byte an attacker could supply that terminates the attribute or starts a new header.

**`userId` is not user input.** It's `account.providerAccountId` from the GitHub OAuth callback (`lib/auth.ts:37`) — a numeric id issued by GitHub and carried in a signed NextAuth JWT. Nobody can set it to an oversized or crafted value without forging the session, at which point the cookie is the least of the problems.

**Oversized wouldn't inject anyway.** A cookie past the browser's ~4KB limit is dropped, not truncated into adjacent headers.

Adding a length guard would be handling a case that cannot occur, so I'm leaving it out.

---

### Closing the review loop

Five rounds, 7 findings:

| Verdict | Count | |
|---|---|---|
| Real, fixed | 3 | stale cookie surviving the param path; `Secure` flag; `Path` scoping |
| Justified, documented in-source | 1 | `HttpOnly` — raised twice; applying it silently breaks the `document.cookie` write path |
| False positive | 2 | param-beats-cookie staleness; this one |
| Re-surfaced resolved | 1 | clustered into the round-4 `HttpOnly` finding after being marked ✅ in round 3 |

Two of the accepted findings came with suggestions that would have caused bugs: the `Secure` fix proposed `NODE_ENV === 'production'`, which breaks self-hosted (production build over plain HTTP), and the param-staleness fix proposed narrowing to `source === 'param'`, which would have reintroduced #498 for first-time visitors. The findings were worth reading every round; the suggestions needed judgment half the time.

Merging on the justifications above.

Issue #506, opened 2026-08-27
Deploy gate and release suite use different concurrency groups — they can tear down each other's fixture run


santthosh:
Two workflows drive the same shared fixtures repo, and each serialises only against itself.

| workflow | repo | concurrency group |
|---|---|---|
| `deploy.yml` → E2E gate | `mergewatch/mergewatch.ai` | **`e2e-fixtures`** |
| `release-suite.yml` | `mergewatch/fixtures` | **`e2e-suite`** |

Different groups, so GitHub happily runs them at the same time.

## Why that is unsafe

Both call `scripts/reset-env.sh`, which does not scope to its own run:

```
#   1. close every open `fixture/*` PR (and delete its remote branch via gh)
#   2. delete every remote `fixture/*` branch
#   3. delete every local branch except main
#   4. return to a clean main reset to the e2e-baseline tag
```

So an overlapping run will close the other run's PRs mid-flight. The victim then sits in `await-reviews.mjs` waiting for check runs on PRs that no longer exist, until it times out — reported as a gate failure with no indication that something else caused it.

## Both workflows already know this hazard

Each carries a comment explaining precisely why it must not overlap — and each solves it only within its own namespace:

**`deploy.yml`**
> Globally serialized, and not per-environment like the deploy jobs. The suite opens real PRs in a SHARED repo and reset-env.sh closes every open PR there — **two concurrent gates would tear down each other's run.**

**`release-suite.yml`**
> Never let two suite runs overlap: **they share fixture branches and would clobber each other's PRs.**

Neither is wrong. Both are incomplete: the resource being protected is *the fixtures repo*, not *this workflow*.

## How likely

Not theoretical. The release suite is dispatchable at any time and also fires on every `v*` tag; the deploy gate fires on every merge to `main`. Cutting a release while a merge is in flight — the most ordinary thing in the world — is enough. A full gate run takes ~20 minutes, so the window is wide.

It has probably not bitten yet only because the release suite is rarely run.

## Fix

One group name shared by both workflows, e.g. `mergewatch-fixtures-repo`, with `cancel-in-progress: false` on both.

GitHub concurrency groups are per-repository, so the two workflows **cannot** share a group directly — they live in different repos. Options:

1. Move the release suite into `mergewatch.ai` alongside the deploy gate, so one group covers both. Also puts it next to the thing it gates.
[...]

PR #507, opened 2026-08-27
feat(ci): release gate — grade the suite, verify the manual set, then cut the tag

### LINKED ISSUE #505: Release gate: run the full fixture suite, verify the manual set, then approve and cut the tag
Today a release is cut and *then* validated. `release-suite.yml` triggers on `push: tags: v*`, so by the time the suite reports, the tag exists and `docker-publish` has already fired on `release: published`. If the suite fails, the release is already out.

This inverts it: **run everything → verify the manual set → approve → cut the tag.**

## Wanted

1. A standalone, manually-dispatched job that runs the **full** fixture suite against a candidate commit.
2. A stage that pauses for a human to work through the **manual** fixtures and record the result.
3. An approval that, once given, cuts the release version.

## What already exists to build on

`mergewatch/fixtures/.github/workflows/release-suite.yml` already applies the suite, waits for reviews, grades deterministically, and tears down. The machinery is there — the trigger and the ordering are wrong, and it does not cover the manual set.

### 1. Teardown must be deferred past human verification

`reset-env.sh` runs as the last step of the suite job and **closes every open PR in the fixtures repo**. Manual fixtures are verified by *looking at the PR* — the comment, the check run, the inline threads. If teardown runs before the human looks, there is nothing to look at.

The run/grade phase and the teardown phase have to be separable, with the verification window between them.

### 2. The two suites do not serialise against each other — and fixing that naively deadlocks merges

| workflow | concurrency group |
|---|---|
| `deploy.yml` E2E gate | `e2e-fixtures` |
| `release-suite.yml` | `e2e-suite` |

**Different groups.** They share one fixtures repo and one `reset-env.sh` that closes every open PR, so a release run and a merge's gate can already clobber each other today. That is a latent bug independent of this ticket.

The obvious fix — one shared group — creates a worse problem here: a release job parked waiting for a human would hold the fixtures lock, and **every merge's E2E gate would queue behind it**. That is exactly the #428 failure (a run parked at a gate held the `deploy-pipeline` lock; one sat ~9.5 hours and GitHub cancelled the next run outright).

So the human-verification window **must not hold the fixtures lock**, which argues for the phase split in (1) being a hard requirement rather than a convenience.

### 3. "All fixtures" needs defining, because most of them cannot fail

| selection | count |
|---|---|
| correctness, total | **80** |
| automated | 52 |
| **automated + graded** | **22** |
| manual | 28 |
| ungraded | 30 |

The deploy gate runs the 22 graded ones. The other **30 automated fixtures run but have no `expect.json`, so they cannot fail** — running them produces PRs, spends LLM budget, and proves nothing. Including them in a release gate would be the coverage illusion this effort has spent months removing.

So: does "run all fixtures" mean all 52 automated (30 of which are theatre), or grade them first? Grading the 30 is arguably the higher-value work and could be a prerequisite.

### 4. Manual verification needs to produce a durable record

28 fixtures, each with a RUNBOOK card listing expected outcomes and failure modes. "I looked at them and they seemed fine" is not a release artifact. The stage should emit a per-fixture pass/fail with the operator's name attached, stored somewhere the next release can compare against.

## Suggested shape

```
workflow_dispatch(candidate_ref)
  │
  ├─ 1. run + grade the automated set        [holds e2e-fixtures lock]
  │     fail fast — a red graded fixture stops here, no human time spent
  │
  ├─ 2. release the lock, leave fixture PRs OPEN
  │
  ├─ 3. manual verification                  [NO lock; environment protection rule]
  │     - renders the 28 RUNBOOK checklists
  │     - human records pass/fail per fixture
  │
  ├─ 4. approve  →  tear down fixtures, cut the tag, publish
  └─ reject      →  tear down fixtures, no tag
```

Step 4 cutting the tag is what makes this a gate rather than a report: nothing reaches `v*` — and therefore nothing reaches `docker-publish` — without passing.

## Acceptance criteria

- [ ] Dispatchable against a candidate commit; **no** `v*` tag exists until approval
- [ ] Runs the full graded set and fails fast before consuming human attention
- [ ] Fixture PRs stay open through verification; teardown happens after the decision, either way
- [ ] The verification window does **not** hold the `e2e-fixtures` lock, so merges keep deploying
- [ ] Both suites share one concurrency group for the phases that *do* touch fixtures
- [ ] Manual results are recorded per fixture, with who and when, not as a single checkbox
- [ ] A rejected candidate tears down cleanly and leaves no tag, no release, no published image
- [ ] The gate reports the same four honest outcomes the deploy gate does — pass / fail / skipped / nothing-selected — never conflating them

## Open question for whoever picks this up

Is grading the 30 ungraded fixtures a prerequisite? A release gate whose headline number counts fixtures that cannot fail is worse than one that honestly says "22 graded, 30 unverified, 28 manual" — which is what the deploy gate already reports today.

santthosh:
Closes #505.

Today a release is cut and **then** validated: `fixtures/release-suite.yml` triggers on `push: tags: v*`, so by the time it reports, the tag exists and `docker-publish` has already fired on `release: published`. A failure arrives after the thing it was meant to prevent.

This inverts it. Dispatch with a version and a candidate ref; **nothing reaches a `v*` tag without a green graded suite AND a human approval.**

## Three jobs, and the split is the design

| job | holds `e2e-fixtures`? | why |
|---|---|---|
| **suite** | ✅ | same group as the deploy gate, so a release and a merge's gate can never drive the shared fixtures repo at once (**#506** — they were in different groups). Tears down in `always()`, *before* any human waits. |
| **verify** | ❌ **deliberately** | parked on an approval while holding that lock would queue every merge's gate behind it. **#428** is what that looks like: one run sat ~9.5 hours and GitHub cancelled the next outright. |
| **release** | — | annotated tag + GitHub release, which is what triggers `docker-publish`. Only reachable through an approved `verify`. |

## A design change from the issue text

#505 proposed leaving fixture PRs **open** through verification — which is precisely what forced the lock question.

It isn't needed. Manual fixtures require a human to **apply and drive** them (reply in a thread, add a reaction, assert Dynamo or dashboard state), not to inspect a PR left lying around — and several have no PR at all (`MODE=dynamo`, `MODE=dashboard`). Tearing down immediately **dissolves** the deadlock rather than managing it. fixtures#1082's `NEEDS_HUMAN_STEP` is what makes the local half workable.

## What it actually verifies

Selection is `--automated --graded`, and an **empty selection is a hard failure** — a gate that runs nothing must not report success.

An ungraded fixture opens a PR, spends real budget, and cannot fail; counting those toward a release headline is the coverage illusion. So the gate reports what is genuinely verified — today **22 graded**, with **27 ungraded** and **31 manual** named in the NOT VERIFIED line. fixtures#1076 raises that number honestly.

## Guarded, not just written

12 tests over the workflow, covering the things that are easy to break by editing it and impossible to notice until a release goes wrong:

- `release` needs `verify`; no `git tag` appears before it
- `suite` shares `e2e-fixtures` with `cancel-in-progress: false`
- `verify` has **no** concurrency group
- selection stays graded-only; empty selection fails
- the tooling checkout is restored **before** grading (otherwise every fixture grades UNGRADED and the gate is green having verified nothing)
- teardown is `always()`
- **no workflow input is interpolated into a `run:` block** — script injection

Verified load-bearing: adding a concurrency group to `verify` fails the test written to prevent it.

```
× does NOT hold that lock while waiting for a human
```

## Also

- Companion PR in fixtures removes the `v*` trigger, which would otherwise re-run the whole suite against the tag this gate just cut, in a different concurrency group.
- The `release-approval` environment has been created with required reviewers.
- `js-yaml` added to `packages/lambda` devDeps (the parser core already uses) with the lockfile updated, so the workflow test can assert structure rather than grep text.

```
build      12/12
typecheck  20/20
test       21/21
```

## Not done

I have **not** dispatched this against a real version — doing so cuts a real tag and publishes real images. It should get its first run with `dry_run: true`, which stops after grading.


santthosh:
All three taken. The critical is a serious catch — **the gate contained the exact failure it exists to prevent.**

## 🔴 Unvalidated `candidate_ref` — real, and the worst kind

`candidate_ref` defaults to `main`, a **mutable** branch, and the release job re-resolved it *after* the approval window:

```
suite    checkout main -> commit A -> graded, passes
human    approves, possibly hours later
release  checkout main -> now commit B -> tags B, never graded
```

A gate whose entire purpose is *"nothing ships unverified"* would have shipped whatever landed during the approval. Worse, it would have looked correct: green suite, human sign-off, tag cut.

Fixed by pinning:

- `suite` resolves HEAD to a SHA **before running anything** and exports it
- `verify` and `release` both check out that SHA, never the input ref
- the tag names the SHA explicitly — `git tag -a … "$SHA"` and `--target "$SHA"` — rather than relying on HEAD being where it was left
- `release` **refuses to tag** if HEAD is not the graded commit; tagging is irreversible, so it verifies rather than assumes the checkout held

## 🟡 Clustered: release notes fence

Taken. The grade summary went into a fixed ` ``` ` fence. A backtick in a fixture name or reason would break out and mangle the note. The fence is now sized to the content — same approach as the cited-code block in #477.

## 🟡 `set -o pipefail` without `-u` in the Grade step

Taken, and correctly scoped: that step was the only one of seven missing `-u`. An unset `$MW_STAGE` would have passed an empty `--stage` rather than failing.

## What I'd note about my own tests

I shipped 12 workflow-invariant tests with this PR and **none of them caught this.** They guarded ordering, locks, teardown and injection — and never asked *which commit gets tagged*. The invariant I most needed was the one I did not think to write.

There are now 6 more, and the important one is verified load-bearing:

```
× checks out the pinned SHA downstream, never the mutable input ref
```

`build 12/12 · typecheck 20/20 · test 21/21` — 18 workflow tests.

santthosh:
Declining this one — `js-yaml@4.3.1` exists, and three independent checks say so.

**1. The registry has it.**

```
$ npm view js-yaml@4.3.1 version dist.tarball
version      = '4.3.1'
dist.tarball = 'https://registry.npmjs.org/js-yaml/-/js-yaml-4.3.1.tgz'
```

**2. The lockfile resolved it, in this PR.**

```yaml
js-yaml:
  specifier: ^4.3.1
  version: 4.3.1
```

**3. CI installed it from that lockfile, on this PR.** `Build & Test` ran `pnpm install --frozen-lockfile` and passed in 1m53s — which is precisely the "fresh environment" the finding predicts would fail. `--frozen-lockfile` is the strictest form: it refuses to resolve anything not already pinned.

**On where the spec came from:** I did not invent it. `packages/core/package.json` has declared `"js-yaml": "^4.3.1"` all along, and the root `package.json` carries a deliberate override:

```json
"js-yaml@>=4.0.0 <4.3.1": "4.3.1"
```

That override exists to force **every** transitive js-yaml up to 4.3.1 — it reads as a security pin. Downgrading `packages/lambda` to `^4.1.0` as suggested would be overridden straight back to 4.3.1 anyway, while making the declared spec disagree with what actually installs.

Matching core's spec was the deliberate choice, and it is the one consistent with the override.

No change. `build 12/12 · typecheck 20/20 · test 21/21`, 18 workflow tests.

santthosh:
**Merging over the red check**, with the override recorded here rather than left implicit.

The blocking finding says `js-yaml@^4.3.1` does not exist. It does:

| check | result |
|---|---|
| `npm view js-yaml@4.3.1` | returns version + tarball |
| this PR's lockfile | `specifier: ^4.3.1 → version: 4.3.1` |
| this PR's CI | `pnpm install --frozen-lockfile` **passed** in 1m53s |

That third row is the finding's own failure scenario — a fresh environment installing from the lockfile — and it is green on this very PR.

`packages/core` has declared `^4.3.1` all along, and the root override `"js-yaml@>=4.0.0 <4.3.1": "4.3.1"` forces every transitive copy up to it. The suggested downgrade would be overridden back to 4.3.1 while making the declared spec disagree with what installs.

This is a deliberate human decision to merge, not an unnoticed red check. The earlier critical on this same PR — the release job re-resolving a mutable ref after the approval window — was **real and is fixed**; this override applies to this finding only.

PR #508, opened 2026-08-27
fix(ci): make the fixtures lock an enforceable invariant, not a convention

### LINKED ISSUE #506: Deploy gate and release suite use different concurrency groups — they can tear down each other's fixture run

santthosh:
Closes #506. Companion: mergewatch/fixtures#1096 — **that one lands first** (this PR's docs describe the fixtures repo as no longer having a suite workflow).

## What was already true, and what wasn't

#505 landed `release-gate.yml` holding **`e2e-fixtures`**, the same group `deploy.yml`'s E2E gate holds. Concurrency groups are per-repository and these two now live in the same repository, so they genuinely serialise. AC-2 (`cancel-in-progress: false`) already held on both. fixtures#1094 separately removed `release-suite.yml`'s `push: tags: v*` trigger, so the gate's own tag no longer re-ran the suite against itself.

fixtures#1096 removes the last driver outside this repo.

What was left is the part that makes it stay fixed: **nothing failed when the groups diverged the first time.** Both workflows carried a comment explaining the hazard, and both solved it inside their own namespace. That is not a fix, it is a convention — and a convention that already lost once.

## The invariant, derived rather than restated

`packages/lambda/src/fixtures-lock.test.ts` parses every workflow, finds each job that checks out `mergewatch/fixtures`, and asserts:

- each holds `e2e-fixtures`
- each is `cancel-in-progress: false` — the loser **waits**; cancelling would strand half-applied fixture branches and open PRs, and the next run's failures would read as product regressions
- exactly **one** group name is used across all workflows — two spellings is the original bug in miniature
- no job in the group names an `environment` — an environment is how approvals and wait timers attach, and holding a shared lock across a human wait is #428, where one run sat ~9.5h and GitHub cancelled the next outright

It first asserts it **found at least two such jobs**. Every other assertion is vacuous over an empty list, so a renamed step or a moved repo would otherwise turn this into a green suite asserting nothing — which is the specific failure this area keeps producing, not a hypothetical.

A third fixtures-driving workflow added later now fails the build instead of failing production months later as an unexplained gate timeout.

## Verified it can fail

Renaming `deploy.yml`'s group to `e2e-suite` — the original bug's exact shape — fails 2 of the 7:

```
× 'deploy.yml' · 'e2e-gate' holds the fixtures lock
× uses exactly one group name across every workflow
  Tests  2 failed | 5 passed (7)
```

Reverted; `deploy.yml`'s group is unchanged in this PR.

## Comments and docs (AC-4)

`deploy.yml` said *"two concurrent gates would tear down each other's run"* — true, and the framing that led to this being solved per-workflow twice. Rewritten to name the **resource**: any two jobs touching that repo collide, not just two gates.

`e2e/RUNBOOK.md` gains a **One suite at a time** section. It had nothing on the gate, the group, or the shared repo at all. It also records the case CI cannot cover: nothing serialises a developer's local `run-suite.sh` against a gate already in flight, so it gives the two commands to check first.

## Checks

`pnpm run build`, `pnpm run typecheck`, `pnpm run test` all green (21/21 tasks, 153 lambda tests incl. the 7 new).

No behaviour change to any workflow — the only YAML edit is a comment.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -304,9 +304,24 @@ jobs:
     runs-on: ubuntu-latest
     needs: deploy-dev
     if: vars.E2E_GATE_ENABLED == 'true' && github.event.inputs.skip_e2e_gate != 'true'
-    # Globally serialized, and not per-environment like the deploy jobs. The
-    # suite opens real PRs in a SHARED repo and reset-env.sh closes every open
-    # PR there — two concurrent gates would tear down each other's run.
+    # This group protects a RESOURCE, not this workflow (#506).
+    #
+    # The resource is the `mergewatch/fixtures` repo. Every suite run opens real
+    # PRs there and tears down with reset-env.sh, which closes every open
+    # `fixture/*` PR — it is not scoped to the run that calls it, and it cannot
+    # be: a runner has no way to tell its own fixture branches from another
+    # run's. So ANY two jobs driving that repo at once destroy each other, not
+    # just two gates.
+    #
+    # `e2e-fixtures` is therefore shared with `release-gate.yml`'s suite job,
+    # and is deliberately global rather than per-environment like the deploy
+    # jobs above. It was originally scoped per-workflow, which left the release
+    # suite serialising against itself in a different repo under a different
+    # name — two locks, one resource, no protection. Anything added later that
+    # checks out that repo belongs in this group too; `fixtures-lock.test.ts`
+    # enforces it.
+    #
+    # Never held across a human wait: see release-gate.yml's `verify`.
     concurrency:
       group: e2e-fixtures
       cancel-in-progress: false
--- a/e2e/RUNBOOK.md
+++ b/e2e/RUNBOOK.md
@@ -121,6 +121,30 @@ For re-runs on the same fixture, you can amend + force-push (cheap) instead of c
 
 ---
 
+## One suite at a time (#506)
+
+`mergewatch/fixtures` is a **single shared mutable resource**, not a workspace per run. `scripts/reset-env.sh` closes every open `fixture/*` PR in it and deletes the branch, and it is not scoped to the run that called it — a runner cannot tell its own fixture branches from anyone else's. Two suites in flight therefore tear down each other's PRs.
+
+The victim does not fail loudly. It waits in `await-reviews.mjs` for check runs on PRs that no longer exist, and reports a timeout with no hint that something else caused it. `await-reviews.mjs` now checks whether its own PRs are still open and names teardown when they are not, but the cheaper fix is not to overlap in the first place.
+
+**In CI this is handled.** Every job that checks out the fixtures repo lives in `mergewatch.ai` and holds the `e2e-fixtures` concurrency group with `cancel-in-progress: false`, so the second one queues:
+
+| workflow | job | when |
+|---|---|---|
+| `deploy.yml` | `e2e-gate` | every merge to `main` — impacted subset |
+| `release-gate.yml` | `suite` | cutting a release — full graded set |
+
+A single group only works because both live in the same repository; GitHub concurrency groups do not span repos. That is why the fixtures repo has no suite workflow of its own any more. If you add another job that drives fixtures, put it here and put it in that group — `packages/lambda/src/fixtures-lock.test.ts` fails the build otherwise.
+
+**Locally it is not.** Nothing stops a local `run-suite.sh` from colliding with a gate run that is already in flight. Before starting one, check:
+
+```bash
+gh run list --repo mergewatch/mergewatch.ai --workflow=deploy.yml --status in_progress
+gh pr list --repo mergewatch/fixtures --state open   # non-empty means a run is live
+```
+
+---
+
 ## Full regression checklist
 
 Run these in order — they cover all current behaviors. ~30 minutes end-to-end.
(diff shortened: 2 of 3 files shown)


santthosh:
## Review response — the dropped critical was real

The verdict on \`6fd6e2a\` was **3/5 advisory**, not clean: one critical was flagged by the orchestrator and then dropped by post-orchestrator filtering, so the comment rendered nothing actionable. Pulled the trail out of \`mergewatch-review-traces-prod\` to see what it was.

| severity | outcome | stage | finding |
|---|---|---|---|
| **critical** | dropped | `fp-i-already-applied` | `fixturesJobs` populated at module load — a detection failure produces no diagnostic |
| warning | dropped | orchestrator | `yaml.load()` / `readFileSync()` errors unhandled |
| warning | dropped | orchestrator | `findFixturesJobs` silently skips malformed YAML |
| warning | dropped | orchestrator | workflow-level group not checked for the environment constraint |
| warning | merged | `fp-c-line-dedup` | module-level side effect runs before the sentinel can guard it |
| warning | merged | `w10-clustering` | `yaml.load()` errors will crash the suite |
| info | merged | `fp-c-line-dedup` | inheritance only checked for `group`, not `cancel-in-progress` |

### The critical was correctly identified and incorrectly dropped

`fp-i-already-applied` fired with *"the suggestion is already implemented at the cited location"*. It matched on the sentinel test **existing**. The finding was that the sentinel **cannot run**.

`findFixturesJobs()` ran at module load to feed `it.each`, and threw on a malformed workflow. In vitest that is a **collection error** — the file never loads, so no assertion executes, including the one whose entire purpose is catching detection failure. You get a stack trace where a sentence about the fixtures lock should be. A test built to stop this area silently checking nothing had that exact hole in it.

Same root as the two `yaml.load` warnings and the malformed-YAML one. Fixed together:

- **The scan is total.** Parse errors are collected, never thrown, and asserted empty by a named test.
- **Collected, not skipped** — for the same reason the sentinel exists. A file that fails to parse yields no jobs, which is indistinguishable from *"this workflow doesn't touch fixtures"*. A third fixtures-driving job could hide behind a YAML typo in its own file while the other two held the count above a `>= 2` floor.
- **The sentinel is now an explicit inventory**, not a count. A count cannot see a job being **added** — which is what #506 was about. The author needs to have read the rule before their job starts driving a shared repo; failing here is how they find out, and the fix is one line.

### Verified both new failure modes

Dropping a malformed `.yml` into `.github/workflows`:

```
× understands every workflow file
  Tests  1 failed | 7 passed (8)
```

One named test, message readable, the other seven still run. Before this commit that took the whole file down as a collection error.

Renaming the checkout target so detection breaks:

```
× finds exactly the jobs known to drive it
AssertionError: expected [ 'release-gate.yml · suite' ]
  to deeply equal [ 'deploy.yml · e2e-gate', …(1) ]
```

Both reverted; neither workflow is modified beyond the comment.

### Two findings need no change

*"workflow-level group not checked for the environment constraint"* and *"inheritance only checked for `group`, not `cancel-in-progress`"* are already handled: `job?.concurrency ?? wf.concurrency` carries the **whole object**, so an inherited group brings its `cancel-in-progress` with it, and the environment check reads the same effective value.

Pushed as `98e52d9`. `build` / `typecheck` / `test` green — 21/21 tasks, 8 tests in this file.

Issue #511, opened 2026-08-27
Release v0.6.0 — manual verification record


santthosh:
> ### ⚠️ Corrected — this record was written before the release actually shipped
>
> `v0.6.0` was **not** cut from `4637f563…`. That first attempt failed in the
> `release` job and was abandoned; the tag was deleted. The release that shipped
> is:
>
> | | |
> |---|---|
> | **released commit** | `47e3112a2cb719ec136061cc037320334be95a0b` |
> | **release gate run** | [33120046602](https://github.com/mergewatch/mergewatch.ai/actions/runs/33120046602) — 22 passed · 0 failed · 0 ungraded · **0 graded but not selected** |
> | **release** | [v0.6.0](https://github.com/mergewatch/mergewatch.ai/releases/tag/v0.6.0), published 2026-08-27T22:24:10Z |
> | **images** | `0.6.0`, `0.6`, sha — [run 33122823134](https://github.com/mergewatch/mergewatch.ai/actions/runs/33122823134). Not `latest`; see #513 |
>
> **The verification below still applies.** The only delta between `4637f56` and
> `47e3112` is `.github/workflows/release-gate.yml` and its test — no product
> code, nothing in any shipped artifact — so nothing the manual set covers
> changed. The graded 22 were re-run in full against `47e3112` regardless.
>
> Everything below describes work done against dev state on 2026-08-26 and is
> left unedited.

---

Record for release gate run [`33038980367`](https://github.com/mergewatch/mergewatch.ai/actions/runs/33038980367), graded commit `4637f563a8787302da0f249cb213e839e9c6953b`.

`e2e/MANUAL-VERIFICATION.md`: *"An unrecorded manual pass is indistinguishable from one that never happened."* This is that record — including, deliberately, what was **not** checked.

## Automated — full graded coverage

```
Selected 22 graded fixture(s)
Suite complete: 22 applied, 0 failed, 0 skipped
22 passed · 0 failed · 0 ungraded · 0 skipped · 0 errored
NOT VERIFIED: 58 correctness fixtures outside this run
              — 31 manual, 27 ungraded, 0 graded but not selected
```

`0 graded but not selected` — nothing gradeable was left out. The honest coverage number is **22 graded fixtures**, not "the suite".

## Manual set — what was actually checked

[...]

Issue #510, opened 2026-08-27
Custom-agent findings bypass W2 verification, so a blocking agent can fail a check on a false claim


santthosh:
> **Rewritten 2026-09-03.** The original framing was wrong in a way that would have produced the wrong fix — it claimed custom agents quote text that isn't in the diff, and proposed verifying quotes against the diff. **The quotes were real.** That check would have passed all three findings. The correction is below; the original text is preserved in the first comment.

A custom agent reported three findings asserting that a **TODO comment had been added**. None of the three files contains `TODO`, `FIXME`, `HACK` or `XXX` anywhere:

```
$ grep -niE "todo|fixme|hack|xxx" scripts/await-reviews.mjs \
    scripts/check-baseline-drift.sh scripts/check-baseline-drift.test.mjs
(no matches)
```

## What was actually false

Not the quotes — **the characterization**.

| finding | quoted text | real? | claim |
|---|---|---|---|
| `await-reviews.mjs:28` | *"Exits 2 if the PRs are torn down underneath it…"* | ✅ the docstring | "a new TODO comment" |
| `check-baseline-drift.sh:3` | *"(mergewatch.ai#509)"* | ✅ line 3 verbatim | "a new TODO comment" |
| `check-baseline-drift.test.mjs:113` | *"no e2e-baseline tag is not this check's problem"* | ✅ a test **name** | "a TODO-like comment" |

Every quote is accurate source text. What is fabricated is calling documentation, an issue reference, and a test title *TODOs*.

This matters for the fix: **verifying that quoted text appears in the diff — the original proposal — would have passed all three.** Judging whether text "is a TODO" is a semantic call, which is exactly what grounding and W2 verification exist to make, and exactly what #385 bypasses.

## Where it comes from

Not a product feature. There is **no TODO detection in the codebase**, and the only built-in prompt text on the subject forbids it — `prompts.ts:511`, under `DO NOT report:`:

> Stale TODOs (TODOs are never findings — they're a separate triage process…)

The source is the **`no-todo` org custom agent** that `mergewatch/fixtures`'s `scripts/seed-org-agent.sh` installs as the E2E-68 prerequisite:

```
prompt:          "Flag any new TODO comment"
severityDefault: critical
scope:           { mode: all }
enabled:         true
```

`scope: all` means it runs on **every** PR in that installation.

## Why nothing caught it

`reviewer.ts:2936` — #385, deliberately:

[...]

PR #512, opened 2026-08-27
fix(release): build the notes before the tag, and stop grep killing the step

santthosh:
`v0.6.0` pushed its tag and then died — [run 33038980367](https://github.com/mergewatch/mergewatch.ai/actions/runs/33038980367). Tag on the remote, no release, and the suite's "refuse to re-cut an existing tag" guard blocks retrying at that version.

## The bug is inverted

```bash
longest=$(printf '%s' "$GRADED" | grep -oE '`+' | awk '...')
```

`grep -o` exits **1 when it matches nothing**. `pipefail` propagates it, Actions runs `bash -e`, and the assignment kills the step.

A backtick-free grade summary is the **normal** case. So the fence guard added in #507 — to survive a fixture name containing a fence — aborted the release in every case *except* the rare one it was written for.

Reproduced exactly:

```
$ bash -e t.sh          # GRADED='22 passed · 0 failed · 0 ungraded'
exit=1
$ bash -e t.sh          # same, one backtick present
REACHED: longest=1
exit=0
```

The workflow had never executed before, so nothing caught it.

## Two changes, and the second matters more

**`|| true` on the grep** — fixes this crash.

**The notes are now built BEFORE the tag is pushed** — fixes the class. The crash was *one* way to fail after the irreversible step; there will be others. Everything that can fail now sits on the near side of `git push origin "$VERSION"`, plus a `test -s /tmp/notes.md` guard so empty notes refuse to tag rather than publishing a blank release.

```
before:  verify HEAD → tag → PUSH → build notes → create release
after:   verify HEAD → build notes → guard → tag → PUSH → create release
```

## The test runs the code, it does not read it

Asserting on the text would not have caught this — **the line looked correct**. So the test lifts the fence computation out of the workflow and executes it under the same `bash -e` + `pipefail` the runner uses:

| case | expect |
|---|---|
| no backticks | exit 0, fence ` ``` ` |
| ` ``` ` present | fence ` ```` ` |
| ` ````` ` present | fence ` `````` ` |
| ordering | notes built before `git push`, `test -s` before `git tag -a` |

Verified it fails without the fix:

```
× sizes the fence when the summary has NO backticks
AssertionError: expected 1 to be +0
```

`+0` is the exit code. 22 tests in the file, all green with the fix.

## Also: workflow renamed to "Release Gate"

Display name only. The file is still `release-gate.yml` and `gh workflow run release-gate.yml` is unchanged.

## Recovery for v0.6.0 — separate from this PR

The tag `v0.6.0` already exists and points at `4637f563…`, **the correct graded SHA**. No release, and docker-publish never fired (it triggers on `release: published`).

The gate's guarantee held: no tag without a graded pass and a human approval, and both happened. What broke was note formatting, after the decision. So `v0.6.0` will be published from the existing tag rather than re-running a 22-fixture suite to re-derive a result we already have. Verification record: #511.

## Checks

`build` / `typecheck` / `test` green — 21/21 tasks.


santthosh:
## Review response — the warning was right

**`release-gate.test.ts:172` — ordering checked against `git tag -a`, not `git push origin`.** Correct, and it's the more interesting kind of gap: the assertions were true, they just weren't asserting the thing this PR is about.

I'd written two pairwise checks:

```
notes written  <  git push origin
test -s        <  git tag -a
```

The invariant that matters is **guard before the irreversible push**. "Guard before `git tag -a`" is a different claim, and a local tag is not the failure mode — `git push origin` is. As the finding says, moving the push above `git tag -a` but below the guard reintroduces the bug with both tests green.

Rather than bolt on the suggested third assertion, I replaced the pairs with one assertion over the **whole sequence**:

```
refusing to tag  →  } > /tmp/notes.md  →  test -s /tmp/notes.md
                 →  git tag -a  →  git push origin  →  gh release create
```

Pairwise checks over an ordering will always leave gaps; asserting the sequence closes this one and the ones nobody has thought of yet. It also names the offending pair on failure instead of reporting two opaque integers.

Verified against the reviewer's exact scenario — push moved above tag:

```
× does everything that can fail before the irreversible push
AssertionError: 'git push origin' must come after 'git tag -a':
  expected 2111 to be greater than 2138
```

21 tests in the file, green. `build` / `typecheck` / `test` green across the workspace.

Pushed as `3360fa5`.

---

The second finding was scored speculative and correctly not surfaced as actionable — no change.

santthosh:
## Review response — declining this one, with evidence

**`release-gate.test.ts:178` — "test uses `set -uo pipefail` but workflow uses `set -eo pipefail`".** Both halves of this are wrong, and the suggested change would make the test less faithful, not more.

### The workflow does not use `set -eo pipefail`

```
$ grep -n "set -" .github/workflows/release-gate.yml | sort -u
79:  set -uo pipefail
96:  set -uo pipefail
105: set -uo pipefail
135: set -uo pipefail
191: set -uo pipefail
285: set -uo pipefail
```

Six run blocks, all `set -uo pipefail`. `set -eo pipefail` appears nowhere in the file. There is also a test in this same file (`shell hygiene`) that fails the build if any run block uses something else.

`-e` does not come from the script. It comes from the runner's default shell, `bash -e {0}` — which is exactly why the test invokes `bash -e -c`. The test reproduces the runner precisely: `bash -e` on the outside, the block's own `set -uo pipefail` on the inside.

### `set -uo pipefail` does not disable an inherited `-e`

The finding's mechanism — that the snippet "does not re-enable `-e` for the subshell" — assumes something turned it off. Nothing did. `set -uo pipefail` is `set -u` plus `set -o pipefail`; it does not touch `-e`.

```
$ bash -e -c 'set -uo pipefail
x=$(echo hi | grep -oE "zzz" | cat)
echo "REACHED — -e was NOT active"'
$ echo $?
1
```

It aborted at the assignment and never reached the `echo`. `-e` is active.

### The test already proved this empirically

When I reverted the `|| true` to check the regression test fires, it failed with:

```
AssertionError: expected 1 to be +0
```

`1` is the exit code the snippet returned. That outcome is only possible **if `-e` is active inside the snippet** — without it, the failed assignment would have been ignored, execution would have continued to `printf "%s" "$fence"`, and the status would have been `0`. The test would have silently passed against the broken workflow.

So the harness is demonstrably faithful; that demonstration is what gives the regression test its value.

### The suggestion would regress it

Changing the header to `set -eo pipefail` **drops `-u`** — which the workflow does have. That moves the test away from the runner's actual environment to match a configuration that does not exist.

No change. Flagging it as a data point for #510: a confidently-worded finding that quotes a configuration the file does not contain.

Issue #516, opened 2026-08-27
"Safe to merge" is asserted without ever reading CI — 5/5 on a PR that did not compile


santthosh:
External evaluation, `catalog-service#103`: MergeWatch returned **🟢 5/5 — Safe to merge — No issues found on changed lines** on a PR with a compile-breaking bug (an undeclared variable). A human caught it before merge.

> A tool whose core promise is catching problems before maintainers have to can still wave through code that literally does not build.

That is fair, and the mechanism is not subtle.

## MergeWatch never looks at CI

A repo-wide search for `statusCheckRollup`, `checks.listForRef`, `getCombinedStatusForRef` and `repos.listCommitStatuses` returns **zero** hits. The only checks API call is a write — `client.ts:436`, `octokit.checks.create`, posting our own check run. `check_run` appears inbound only as a trigger, and only for our own check (`webhook.ts:474`).

The verdict's entire input is `reviewer.ts:231`:

```ts
`${resolvedPrompt}\n\n${INTENT_CLAIMS_DIRECTIVE}\n\n--- PR Context ---\n${contextBlock}\n\n--- Diff ---\n${diff}`
```

PR metadata plus the filtered diff, plus files an agent asks for. **No build output, no test results, no CI status.** An undeclared variable is precisely what a typecheck catches and an LLM reading a diff hunk plausibly does not.

## The headline also overclaims in a second way

`reviewer.ts:2629`:

```ts
if (noActionItems) {
  return { mergeScore: 5,
    mergeScoreReason: filteredFindings.length === 0
      ? 'No issues found on changed lines.' : ...
```

That fires when zero action findings survive **every post-orchestrator filter** — grounding, W2 verification, line proximity, dedup, clustering. It does not mean the agents found nothing. "No issues found" and "everything we found was filtered out" render identically, and #510 is an open case of filters dropping a real finding.

So the strongest words in the product — `Safe to merge`, `No issues found` — sit on the narrowest evidence.

## The rule already exists; it just is not applied to unknowns outside the diff

`docs/feat/20260822-01-context-architecture.design.md`:

> Honest coverage + score clamping — whatever is dropped must be stated, and the score must not overclaim. **A 5/5 on a partially-reviewed diff is worse than today's hard failure**, because a hard failure at least signals something went wrong.

[...]

Issue #382, opened 2026-08-19
Non-deterministic post-orchestrator suppression eats legitimate findings (same-run A/B evidence across 4 fixtures)


santthosh:
## Summary
On the 2026-08-19 full-suite run (fixtures PRs #411–#467, sterilized bait, all infra healthy), the dedup/quality-filter layer **non-deterministically suppressed findings that identical or near-identical stimuli surfaced minutes apart in the same run**. This is the successor to #375 with the harness confound removed — these are single-phase fixtures, and the A/B contrast is within one run.

## A/B evidence (strongest first)
1. **Identical-bait triplet, one arm suppressed.** Fixtures 35/39/40 share the same unauthenticated-admin-endpoint bait. [#441](https://github.com/mergewatch/fixtures/pull/441) → 2/5 inline Critical; [#446](https://github.com/mergewatch/fixtures/pull/446) → 1/5 inline Critical; but [#445](https://github.com/mergewatch/fixtures/pull/445) → 🟢 5/5, zero findings, **"Suppressed: 3 findings removed by dedup & quality filters"** — and the summary prose itself concedes the endpoint "lacks … authentication/authorization checks … all of which are critical." Emitted, described, then filtered.
2. **Custom-agent findings filtered.** [#455](https://github.com/mergewatch/fixtures/pull/455) (68-org-custom-agents): no TODO finding at any severity; footer **"Suppressed: 2"** — exactly the two planted `// TODO`s. Custom-agent findings appear not to be filter-exempt (or the org-critical/repo-info precedence resolves to info and dies in the filter).
3. **Convention findings flipped run-over-run.** [#465](https://github.com/mergewatch/fixtures/pull/465) (80a): 5/5 all-clear with **"Suppressed: 7"**, while the review's own mermaid diagram lists all four violations ("Violation: var declaration", "loose equality ==", "any type", "console.log call"). Last run the same fixture cited the var rule at 2/5 ([#405](https://github.com/mergewatch/fixtures/pull/405)). Sibling 80b ([#466](https://github.com/mergewatch/fixtures/pull/466)) cited its EARLY-RULE correctly in the same run — same subsystem, inconsistent outcome.
[...]

PR #530, opened 2026-09-02
fix(core): close our inline threads when a finding is withdrawn

### LINKED ISSUE #526: Mergewatch leaves redacted “critical” findings active, causing inaccurate review status
**Description**
We’re seeing a recurring problem where Mergewatch posts review findings marked as **critical**, then later redacts or withdraws those findings, but the original items still appear to remain active in the PR’s overall review state.

As a result:

- the PR continues to reflect unresolved critical review noise even after those comments have been redacted
- the overall review score/status remains below a full pass (for example, less than 5/5)
- authors and reviewers are left unsure whether there is still real blocking feedback to address

This creates a lot of churn on larger PRs because the review summary no longer matches the current visible feedback.

**Expected behavior**
If Mergewatch redacts, withdraws, or otherwise invalidates a previously-posted critical finding, that finding should no longer count against the PR’s active review state or overall score.

**Actual behavior**
A critical item is posted, later redacted, but still appears to affect the PR’s aggregate review result/status.

**Impact**
- adds review noise
- makes it harder to tell whether a PR is actually ready
- can leave authors stuck with an apparently incomplete/pending review despite no remaining actionable critical comments
- reduces trust in the review summary

**Example scenario**
1. Mergewatch posts one or more “critical” findings on a PR
2. Those findings are later redacted
3. The PR still does not return to a clean/passing overall review state
4. The review summary still appears degraded (for example, below 5/5)

**Request**
Please ensure that redacted or withdrawn findings are fully removed from the active review calculation and no longer count toward the PR’s overall status/score.

If helpful, I can provide a specific PR example where this occurred.

santthosh:
Closes #526. Phase 2 of 2 — Phase 1 (#529) landed as `6beb8d4`. Milestone **v0.6.1**.

## The other half of the report

A withdrawn finding left its inline comment open **forever**:

- `pulls.deleteReviewComment` — does not exist in the codebase
- `minimizeComment` — does not exist
- `resolveReviewThread` — exists, but called from exactly one place: an explicit human `/resolve` reply

And a re-review *deliberately skips re-posting* a finding it still holds, so the original inline comment is the **only** artifact. The PR kept showing unresolved blocking feedback for something the review itself had already dropped — the reporter's "authors and reviewers are left unsure whether there is still real blocking feedback".

## What it does

`resolveWithdrawnFindingThreads` walks our own threads and closes the ones the current review no longer raises, **replying first** so the reason is discoverable rather than the thread silently vanishing.

## Three restrictions, all in the under-resolve direction

Leaving a stale thread open is a nuisance. Closing someone's live conversation is not.

| | |
|---|---|
| **Only threads where every comment is ours** | one reply from anyone else and we leave it alone |
| **Only when identity is provable** | no `selfLogin`, no resolving — the same guard `dismissStaleReviews` uses (#418) |
| **Matched on path + title, never line** | GitHub shifts a thread's line as the PR moves; a line-sensitive key would silently stop matching after a push and could resolve a **live** finding |

A title that genuinely recurs twice in one file keeps **both** threads open — the safe way to be wrong.

It never throws, and one stuck thread does not abandon the rest.

## A test that earned its place immediately

My first fixture hand-wrote the comment body as `### 🔴 title`. The real emitter produces `**🔴 title**`, which is what `extractInlineCommentTitle` matches.

**Every assertion still passed.** An unmatched title yields an empty string, which yields a key that is never in `activeKeys`, which means "withdrawn" — so the block was testing a format the product never produces, and would have gone green forever.

Two changes came out of that:

- the fixture now builds its body with **`buildInlineComments`**, the real emitter, so it cannot drift from what ships
- a `the fixture body is one the extractor actually understands` test guards the guard — without it the whole block passes vacuously

That is the same shape as #516 and fixtures#1076: something that looks like coverage and cannot fail.

## Tests

11 new, including the ones that must **not** fire — a human reply in the thread, a thread someone else started, a comment of ours without the marker, an already-resolved thread, and unknown identity.

```
Test Files  1 passed (1)
     Tests  126 passed (126)
```

`build` / `typecheck` / `test` green — 21/21 tasks.

## What #526 does not include, and why

**The score floor stays.** `reviewer.ts:2620`/`:2679` hold the score at 3 when the orchestrator raised criticals and filtering dropped them all — #382/#385's guard, which exists because silent filtering once produced a 5/5 on a self-admitted SQL injection.

**Withdrawn vs. Resolved in the rendered output** is v0.7.0 shaped. A withdrawn finding currently renders as "✅ Resolved on this commit", indistinguishable from one the author fixed (`review-delta.ts:108`). Tracked as P13 in `docs/review-quality-plan.md:158`.

### DIFF
--- a/packages/core/src/github/client.test.ts
+++ b/packages/core/src/github/client.test.ts
@@ -1146,6 +1146,40 @@ describe('withdrawn finding threads are closed (#526)', () => {
     expect(withdrawnThreadKey('src/app.ts', '  SQL Injection In The Query Builder  ')).toBe(KEY);
   });
 
+  it('leaves a thread alone when its comment list is truncated', async () => {
+    // >100 comments means the "ours alone" check answers a question about the
+    // first page, not the thread. A human reply at position 101 would be
+    // invisible and the thread resolved out from under them.
+    const t = thread();
+    (t.comments as Record<string, unknown>).pageInfo = { hasNextPage: true };
+    const { octokit, mutations } = stub([t]);
+
+    expect(await resolveWithdrawnFindingThreads(octokit, 'o', 'r', 7, new Set(), ours)).toBe(0);
+    expect(mutations).toEqual([]);
+  });
+
+  it('still resolves when the comment list is explicitly NOT truncated', async () => {
+    const t = thread();
+    (t.comments as Record<string, unknown>).pageInfo = { hasNextPage: false };
+    const { octokit, mutations } = stub([t]);
+
+    expect(await resolveWithdrawnFindingThreads(octokit, 'o', 'r', 7, new Set(), ours)).toBe(1);
+    expect(mutations).toEqual(['T1']);
+  });
+
+  it('leaves a thread alone when the title cannot be parsed', async () => {
+    // An empty title yields a key in no active set, which reads as
+    // "withdrawn" — resolving a thread we failed to identify. This is the
+    // exact mistake the original fixture made by accident.
+    const t = thread({
+      comments: { nodes: [{ databaseId: 40, body: `${bodyFor('x').split('**')[0]}### 🔴 wrong format`, author: { login: ours } }] },
+    });
+    const { octokit, mutations } = stub([t]);
+
+    expect(await resolveWithdrawnFindingThreads(octokit, 'o', 'r', 7, new Set(), ours)).toBe(0);
+    expect(mutations).toEqual([]);
+  });
+
   it('survives a lookup failure without throwing', async () => {
     const octokit = { graphql: vi.fn(async () => { throw new Error('boom'); }) } as unknown as Octokit;
     expect(await resolveWithdrawnFindingThreads(octokit, 'o', 'r', 7, new Set(), ours)).toBe(0);
--- a/packages/core/src/github/client.test.ts
+++ b/packages/core/src/github/client.test.ts
@@ -9,6 +9,8 @@ import {
   postReviewComment,
   updateReviewComment,
   createCheckRun,
+  resolveWithdrawnFindingThreads,
+  withdrawnThreadKey,
   enforceCommentBodyLimit,
   MAX_COMMENT_BODY,
   parseRepoConfigYaml,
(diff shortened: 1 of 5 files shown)


santthosh:
## Review response — all three findings were right

This PR claimed a safety invariant:

> Three restrictions, all in the **under-resolve** direction — leaving a stale thread open is a nuisance, closing someone's live conversation is not.

The review found **three places the code did not honour it.** Each silently produced "withdrawn" and resolved a thread it should not have touched. Fixed in `64e47dc`.

### 1 — Truncated comment list (critical, correct)

`comments(first: 100)` with no pagination check meant `nodes.some(c => …!== want)` was answering a question about **the first page**, not the thread. A human reply at position 101 was invisible, and the thread would be resolved out from under them.

Now requests `pageInfo { hasNextPage }` and **skips any thread it cannot see all of** — the suggestion was to warn; skipping is stronger and matches the rest of the function.

### 2 — Unparseable title (warning, and the sharpest of the three)

`extractInlineCommentTitle` returns `''` when the body does not match, and an empty key is in no active set — so **a comment we failed to identify read as withdrawn.**

This is the *exact* mistake my own test fixture made by accident, arriving back as a product bug. The fixture used `### 🔴 title` where the emitter produces `**🔴 title**`; the title came out empty and the assertion passed anyway. I fixed the fixture and did not ask whether the same input could reach production code. It can — any inline comment predating the current body format. Now skipped.

### 3 — Degenerate finding keys (critical, correct)

`result.findings as Array<{ file: string; title: string }>` was an unchecked cast. An undefined field yields `undefined::…`, which matches no real thread — so the **live** thread looks withdrawn and gets resolved. Now filtered to entries with a usable `file` and `title`, in both runtimes as suggested.

Dropping a malformed finding leaves its thread **open**, which is the safe direction.

### Also

The thread list itself can truncate at 100. That one is genuinely under-resolve and harmless, but it now logs — "why is that thread still open" deserves an answer.

## Verification

Three new tests. Verified they fail without the guards:

```
× leaves a thread alone when its comment list is truncated
× leaves a thread alone when the title cannot be parsed
  Tests  2 failed | 127 passed (129)
```

129 passing now; `build` / `typecheck` / `test` green across the workspace.

The common thread in all three is the rule already applied to `selfLogin` and to human replies: **when identity cannot be established, do nothing.** I stated that invariant in the PR description and then did not apply it to truncation, to unparseable bodies, or to malformed findings.

santthosh:
## Review response — right, and fixed a different way than suggested

**Partial failure leaves a misleading comment on an open thread.** Correct. Reply-then-resolve in one try/catch means a thrown `resolveReviewThread` leaves the thread **open**, carrying a comment that says *"— resolving"*. A comment asserting a state that does not exist, on the very thread it is wrong about. And the single catch made "both failed" indistinguishable from "note posted, thread still open".

### Why not the suggested ordering

The suggestion was to resolve first, so a failed reply is merely cosmetic. That is sound reasoning, and it rests on **replying to a resolved thread not re-opening it** — which I could not confirm without assuming GitHub's behaviour. Given this PR is entirely about not resolving threads on unverified assumptions, building its own correctness on one seemed like the wrong trade.

### What landed instead

**The note now states only what is already true when it is written:**

> MergeWatch is no longer raising this finding on the latest commit.

That sentence holds whether the resolve succeeds, fails, or never runs. The inconsistent state is removed **by construction** rather than by ordering — no dependency on API semantics either way.

Plus the two operations get their own catches, and the resolve failure logs whether the note landed:

```
Failed to resolve withdrawn thread T1 (o/r#7) — note was posted: …
```

That was the second half of the finding and it was right.

### A behaviour change that fell out

A failed note no longer abandons the thread — the note is a courtesy, resolving is the point. That changed semantics an existing test asserted (`one stuck thread does not abandon the rest` was making the *reply* fail), so it now covers a stuck **resolve**, which is the case its name describes.

### Verification

```
Tests  131 passed (131)
```

Two new tests: the note never contains "resolving"/"resolved" when the resolve throws, and a failed note still resolves. `build` / `typecheck` / `test` green.

Pushed as `d7c0eda`.

Issue #558, opened 2026-09-04
The "Re-run" button does nothing — GitHub fires check_suite.rerequested and the webhook only handles check_run


santthosh:
## What happens

Clicking **Re-run** on the `MergeWatch Review` check does nothing. No new review, no check-run update, no error. The check run keeps the original review's timestamps:

```
id=101112907899  completed/failure  started=17:24:34Z  completed=17:25:06Z
```

## Cause

GitHub delivers **`check_suite.rerequested`**, and the webhook's dispatch switch has no case for it, so it falls to the default branch and is discarded:

```
18:05:22  Ignoring unhandled event type: check_suite
18:19:02  Ignoring unhandled event type: check_suite
18:23:51  Ignoring unhandled event type: check_suite
```

Three deliveries, matching three Re-run clicks.

The App **is** subscribed and the events **do** arrive. What is missing is a handler:

```ts
switch (githubEvent) {
  case "pull_request": ...
  case "issue_comment": ...
  case "pull_request_review_comment": ...
  case "check_run":  await handleCheckRunEvent(payload as CheckRunEvent); break;
  case "installation": ...
  case "marketplace_purchase": ...
  default:
    console.log(`Ignoring unhandled event type: ${githubEvent}`);
}
```

`handleCheckRunEvent` is complete and correct — including #416's stage-aware name matching so a non-prod App recognises its own `MergeWatch Review (dev)` check. It handles `check_run.rerequested`. It is simply not the event GitHub sends for this affordance.

No `check_run` event has been observed in the logs at all, over three days, while `check_suite` arrives reliably.

## Impact

- The **Re-run** button is inert for every user on every PR in production — a visible affordance that silently does nothing.
- `12-rerun-check` cannot pass, and has presumably never passed: it is a `MANUAL_ONLY` fixture whose entire procedure is clicking this button, and nothing else exercises the path.
- Re-review still works via a new commit or `@mergewatch review`, so users have a workaround — they just have no way to know the button is not one.

## Fix

- Add a `check_suite` case that handles `action === 'rerequested'`, resolving the PR from `check_suite.pull_requests[0]` and dispatching the same re-review as `pull_request.synchronize`.
- Keep the existing `check_run` case: GitHub sends `check_run.rerequested` when a *single* check run is re-run, and `check_suite.rerequested` for "Re-run all checks". Both are legitimate entry points and the two payload shapes differ.
[...]

Issue #548, opened 2026-09-04
A missing GitHub permission should be loud: surface it, don't warn-and-continue


santthosh:
Follow-up to #545, which is closed — the permission is granted and both dependent features are verified working. This covers the hardening that #545 identified but did not implement.

## Why this is worth doing separately

#545 was a missing `Contents: write`. The permission itself was a two-minute fix. What made it expensive was that **nothing said so**:

- `resolveWithdrawnFindingThreads` catches the FORBIDDEN and `console.warn`s (`client.ts:122`, `:132`), so #526 shipped, ran on every review for a day, resolved nothing, and reported success.
- FP-F calls `resolveReviewThread` with no catch, so the whole inline-reply job threw. The user's `resolved` produced no resolution, no acknowledgement, and no explanation.
- Its unit tests mock Octokit, so they passed throughout and could not distinguish "resolved the thread" from "was forbidden".

It took driving a fixture by hand against a deployed stage to find. A permission error is not a transient failure and should not be logged like one.

## Scope

**1. Distinguish permission failures from transient ones.**
A `FORBIDDEN` / `Resource not accessible by integration` is deterministic: it will fail identically on every subsequent call. Log it at `error`, name the permission if derivable, and say which feature is degraded. Once per run, not once per thread — the batch-summary pattern already used for `appendRejectReason` failures fits.

**2. Give FP-F a real failure path.**
If the thread cannot be resolved, the user's `resolved` should still be acknowledged in-thread and the disposition still recorded. Losing the whole job because the tidy-up step failed inverts the priorities — the signal matters more than the thread state.

**3. A capability check.**
On startup or first use, verify the App holds the permissions its enabled features need, and report a missing one plainly, once. Two features were silently non-functional and the only signal was a `warn` nobody had reason to read.

**4. Consider a live-API test for permission-dependent paths.**
Not necessarily a fixture — even a single authenticated smoke check against the real API would have caught this, where a mocked unit test structurally cannot. This is the third instance in one week of "the code runs, logs success, and achieves nothing" (#510, #526, #544); the common thread is that the assertion sat one level below where the bug was.

## Acceptance

- [ ] A permission-denied error is logged at `error`, once per run, naming the degraded feature
- [ ] It is textually distinguishable from a transient/network failure
- [ ] FP-F records the disposition and answers the user even when resolution fails
- [ ] A missing permission for an enabled feature is reported once, plainly, rather than per-occurrence
[...]

Issue #545, opened 2026-09-04
The App is FORBIDDEN from resolving review threads — FP-F inline resolve errors out and #526 is silently inert


santthosh:
## What happens

`resolveReviewThread` returns **FORBIDDEN** for our GitHub App, on a thread **the App itself created**:

```
Inline reply failed for <repo>#1707: GraphqlResponseError
 - Resource not accessible by integration
  errors: [ { type: 'FORBIDDEN', message: 'Resource not accessible by integration' } ]
  data: { resolveReviewThread: null }
  request: mutation ResolveThread($threadId: ID!) { resolveReviewThread(input: { threadId: $threadId }) { thread { id isResolved } } }
    at resolveReviewThread (packages/core/src/github/client.ts:1123)
    at handleInlineReply    (packages/core/src/agents/inline-reply.ts:536)
```

Reproduced by replying `resolved` in a MergeWatch inline finding thread on the production stage, while driving `35-inline-resolve` during v0.6.1 manual verification.

## Two features depend on this, and they fail differently

**1. FP-F inline resolve (E2E-35) — fails loudly, and the user's action is lost.**
`handleInlineReply` calls `resolveReviewThread` directly with no catch, so the whole inline-reply job throws. A user typing `resolved` gets no resolution, no bot acknowledgement, and no explanation. The thread stays open.

**2. `resolveWithdrawnFindingThreads` (#526) — fails silently.**
That path wraps each resolve in `try/catch` + `console.warn`:

```ts
} catch (err) {
  console.warn('Failed to note withdrawal on thread %s (%s/%s#%d):', thread.id, owner, repo, prNumber, err);
}
```

So the FORBIDDEN is swallowed as a warning and the feature does nothing. **#526 shipped today** — unit tests green, review clean, E2E gate green — and it cannot work in production, because nothing it depends on is permitted. Its tests mock the Octokit call, so they prove the logic and say nothing about whether the mutation is allowed.

This is the same shape as #510's near-miss: the code runs, logs success, and achieves nothing, because the one thing that would have revealed it was never exercised against the real API.

## Cause (to confirm)

The App is refused a mutation on its own thread, so this is a permission/capability gap rather than an ownership one. Candidates, in order:

- The App's granted permissions lack whatever `resolveReviewThread` requires (it is not covered by the same grant that allows posting reviews and inline comments).
[...]

Issue #544, opened 2026-09-04
Inline replies are silently ignored on every non-prod stage — the stage is never passed to the marker check


santthosh:
## What happens

On a non-prod stage, replying in a MergeWatch inline thread does nothing. `/mergewatch reject`, `/resolve`, and ordinary threaded conversation are all dropped. No reply, no reaction, no error — the log says only:

```
Inline reply skipped for <repo>#1709 (reply=3929798322, cost=$0.0000)
```

## Proof

Found while driving `40-mergewatch-reject` by hand during v0.6.1 manual verification. Both Apps are installed on the fixtures repo, so the same PR carries two inline finding threads — one per stage. I posted the **identical** command in each, a minute apart:

| Thread replied to | Result |
|---|---|
| dev App's thread | `Inline reply skipped` — nothing recorded, no reaction, no footer |
| prod App's thread | `Inline reply rejected` · `[fb-d] recorded 2 /mergewatch rejects (category=style-disagreement)` · finding comment edited with `✅ Marked rejected (style-disagreement)` |

Same command, same PR, same code build. Only the stage of the thread differed.

## Cause

`packages/lambda/src/handlers/review-agent.ts:260` builds the inline-reply context without a stage:

```ts
const result = await handleInlineReply(
  {
    owner,
    repo,
    prNumber,
    replyCommentId: inlineReplyCommentId,
    conventions: conventionsResult?.content,
    // <- no `stage`
  },
  { octokit, llm, lightModelId: ... },
);
```

`handleInlineReply` then guards the thread root with `inlineMarker(ctx.stage)` (`packages/core/src/agents/inline-reply.ts:509`), and `stage.ts:51`:

```ts
function isProd(stage?: Stage): boolean {
  return !stage || stage.trim() === '' || stage.trim().toLowerCase() === 'prod';
}
```

An absent stage **is prod**. So the marker check always looks for the prod marker `<!-- mergewatch-inline -->`, while a dev thread root carries `<!-- mergewatch-inline:dev -->`. That is not a substring match — the `:dev` sits before the closing ` -->` — so the guard concludes "thread root is not a MergeWatch comment" and returns `skipped`.

The App is correctly reading a thread it wrote itself, and rejecting it as someone else's.

## Why it stayed hidden

Two reinforcing reasons:

[...]

PR #559, opened 2026-09-04
fix(release): grant actions:write so the gate can dispatch docker-publish

santthosh:
Found by cutting v0.6.1.

## What happened

The release gate tagged `v0.6.1`, created the GitHub release, and then failed:

```
could not create workflow dispatch event: HTTP 403: Resource not accessible by integration
https://api.github.com/repos/mergewatch/mergewatch.ai/actions/workflows/244437205/dispatches
```

**The release exists; the images did not.** They were published manually afterwards.

## Cause

```yaml
permissions:
  contents: write
```

Dispatching a workflow is an **Actions API write**, not a contents write, and `GITHUB_TOKEN` grants exactly what the permissions block lists. So `gh workflow run docker-publish.yml` was forbidden.

## The guard worked

This is the failure #513/#514 anticipated. Rather than reporting a green release with no images, the step failed with:

```
::error::docker-publish never started for $VERSION — the release exists without images
```

That is the whole point of that guard, and it is why this was caught in the run rather than discovered later by someone pulling a missing tag.

## Tests

Two assertions on the workflow's declared permissions — the only level at which this failure is visible, since nothing about the dispatch step's code is wrong:

- `actions: write` is granted (**fails without this change** — verified by reverting)
- `contents: write` is still granted, so a future fix cannot replace it rather than add to it

Full forced gate green — build 12/12, typecheck 20/20, test 21/21, all `Cached: 0`.

## Fourth of a kind this week

| | Code | Missing | Symptom |
|---|---|---|---|
| #544 | correct | `stage` not passed | replies silently ignored |
| #545 | correct | `Contents: write` | resolve silently forbidden |
| #558 | correct | `check_suite` handler | Re-run button inert |
| this | correct | `actions: write` | release ships without images |

Every one: implementation right, an out-of-band grant or wiring missing, and unit tests green because the missing piece is not in the code under test. This one was caught only because someone had already written a guard that refused to call it a success. That argues for #548's capability check covering **declared workflow permissions** too, not just App permissions and webhook subscriptions.

### DIFF
--- a/.github/workflows/release-gate.yml
+++ b/.github/workflows/release-gate.yml
@@ -53,6 +53,12 @@ on:
 
 permissions:
   contents: write
+  # `actions: write` is what lets the release job DISPATCH docker-publish.
+  # Without it the dispatch fails 403 "Resource not accessible by integration"
+  # and the release exists with no images — which is exactly what v0.6.1 did.
+  # It is not covered by `contents: write`: dispatching a workflow is an Actions
+  # API write, and GITHUB_TOKEN grants only what is declared here.
+  actions: write
 
 jobs:
   suite:
--- a/packages/lambda/src/docker-publish.test.ts
+++ b/packages/lambda/src/docker-publish.test.ts
@@ -63,6 +63,27 @@ describe('docker-publish — only a release ships images', () => {
 });
 
 describe('release gate — the release actually ships images', () => {
+  // v0.6.1: the gate tagged, created the release, then failed dispatching
+  // docker-publish with
+  //   HTTP 403: Resource not accessible by integration
+  //     .../actions/workflows/<id>/dispatches
+  // because the workflow declared only `contents: write`. Dispatching a
+  // workflow is an ACTIONS API write, and GITHUB_TOKEN grants exactly what the
+  // permissions block lists — so the release shipped with no images and a human
+  // had to publish them by hand.
+  //
+  // Asserted here rather than in the dispatch step because the failure is a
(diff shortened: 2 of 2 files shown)


santthosh:
## Merged, live, and deliberately not deployed

This change needs no prod deploy, and one was not run. Recording that here so the gap between `main` and prod's last deploy SHA is not mistaken later for an unfinished ship.

The diff is two files:

- `.github/workflows/release-gate.yml` — GitHub Actions reads workflows **from the repository**, not from deployed infrastructure. The `actions: write` grant took effect the moment this merged and will apply at the next release cut.
- `packages/lambda/src/docker-publish.test.ts` — a test file, not in the Lambda bundle.

No runtime code changed. A deploy would have built and shipped a byte-identical Lambda.

### About the gate failure on this merge

The E2E gate failed on `29-cluster` and prod was skipped. That failure is **unrelated** to this PR — a +27-line change to a YAML permissions block and a test cannot alter review behaviour. The review on that fixture worked correctly and produced 4 criticals; they were four *distinct* defects, so W10 had nothing to merge and the clustered-audit-trail string never rendered. Clustering was healthy in the same run (`merged 1/3/3/1` on other fixtures).

Filed as mergewatch/fixtures#1858: the fixture asserts a product literal whose *presence* is gated on a model outcome, while its own note claims it avoids model judgement.

The gate was **not** re-run and **not** skipped. Re-running would have cost a full graded suite to prove something the diff already proves, and "re-run on red" is the habit that makes a gate worthless. This fix will ride out with the next merge's deploy, behind a real gate run.

PR #562, opened 2026-09-04
fix(core): handle check_suite.rerequested so the Re-run button works (#558)

### LINKED ISSUE #558: The "Re-run" button does nothing — GitHub fires check_suite.rerequested and the webhook only handles check_run

santthosh:
Closes #558.

## The bug

Clicking **Re-run** on the `MergeWatch Review` check did nothing — in production, for every user, on every PR. No review, no error, and the only trace was:

```
Ignoring unhandled event type: check_suite
```

GitHub fires **`check_suite.rerequested`** for that button. The webhook handled only `check_run.rerequested`, so the event hit the dispatch switch's `default` branch and was discarded. `handleCheckRunEvent` was complete and correct the entire time — including #416's stage-aware name matching — it was simply never reached.

Three clicks produced three `check_suite` deliveries on the prod webhook and **no** `check_run` event in three days of logs.

## The fix

- `case "check_suite"` handling `rerequested`, resolving the PR from `check_suite.pull_requests[0]`.
- `check_run` is **kept** — GitHub sends it when a *single* check run is re-run, a different affordance with a different payload shape.
- The shared tail (PR fetch → classification → existing-comment lookup → enqueue) is **extracted rather than duplicated**, so the two entry points cannot drift into producing different review jobs.

## Ownership, and why it is not `app.id`

A suite has no `name`, so `isMergeWatchCheckRun`'s rule cannot be applied to the payload. Its own comment explains why `app.id` is not the identity check:

> We match by name (stable across deploys) since `check_run.app.id` requires knowing our GitHub App ID at runtime.

So the same rule is applied one level down: **our stage's check run must exist on the suite's head SHA.** That survives an App rename, keeps dev from acting on prod's suite, and costs one API call on an event that only fires when a human clicks a button.

It **fails closed**. A missed re-run is recoverable by pushing a commit; re-reviewing another tool's suite spends real money on work nobody asked for.

## On fan-out

The acceptance criteria asked that a suite re-run not fan out into duplicate reviews. That **resolves by construction** — one suite event yields one job however many check runs the suite holds, and there is a test asserting exactly that.

Separately, `check_run.rerequested` and `check_suite.rerequested` do not co-fire today (three clicks → three `check_suite`, zero `check_run`), so handling both does not double-enqueue. The enqueue path has no dedup, so I have **not** built speculative dedup for a case with no evidence — flagging it here rather than hiding it.

## Tests

Seven added. **Two fail without the handler** (the positive cases); the other five are negative guards — ignore a non-ours suite, fail closed on an ownership-check error, ignore `requested`/`completed` (which fire on every push and would double every review), ignore a suite with no PR, ignore a missing installation id. Those pass trivially against a no-op baseline by design: they exist to stop the *new* code over-firing, not to catch the original bug.

Full forced gate green — build 12/12, typecheck 20/20, test 21/21, all `Cached: 0`.

## Provenance

Found by driving `12-rerun-check` by hand during v0.6.1's manual verification — the fixture whose entire procedure is clicking this button, and which had therefore never passed. The gate cannot catch this class: it is an event the fixture suite has no way to emit.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -292,6 +292,7 @@ export type {
   MarketplacePurchaseEvent,
   MarketplacePurchase,
   CheckRunEvent,
+  CheckSuiteEvent,
   CheckRunPullRequestRef,
   WebhookEvent,
   ReviewMode,
--- a/packages/core/src/types/github.ts
+++ b/packages/core/src/types/github.ts
@@ -281,6 +281,30 @@ export interface CheckRunEvent {
   sender: GitHubUser;
 }
 
+/**
+ * `check_suite` — what GitHub actually fires for the "Re-run" button.
+ *
+ * #558: the webhook handled `check_run.rerequested` and GitHub sends
+ * `check_suite.rerequested`, so the button silently did nothing in production
+ * for every user. A suite has no `name`, which is why ownership cannot be
+ * decided the way `isMergeWatchCheckRun` decides it.
+ */
+export interface CheckSuiteEvent {
+  action: "completed" | "requested" | "rerequested";
+  check_suite: {
+    id: number;
+    head_sha: string;
+    status: string | null;
+    conclusion: string | null;
+    /** The App the suite belongs to. Present, but see #558 on why it is not the identity check. */
+    app?: { id: number; slug: string; name: string };
+    pull_requests: CheckRunPullRequestRef[];
+  };
+  repository: GitHubRepository;
+  installation?: { id: number };
+  sender: GitHubUser;
+}
+
 // ---------------------------------------------------------------------------
 // Discriminated union for routing
 // ---------------------------------------------------------------------------
--- a/packages/lambda/src/handlers/webhook.test.ts
+++ b/packages/lambda/src/handlers/webhook.test.ts
@@ -643,6 +643,16 @@ describe('handler — check_suite.rerequested (#558)', () => {
     expect(mockEnqueue).not.toHaveBeenCalled();
   });
 
+  it('resolves the installation client once, not once per step', async () => {
+    // Review finding on #562: the ownership check and the enqueue each resolved
+    // their own client, so every Re-run cost two token exchanges for the same
+    // installation — and gave the second one a chance to fail after ownership
+    // had already passed.
+    await handler(makeCheckSuiteApiEvent(JSON.stringify(makeCheckSuiteEvent())));
+    expect(mockEnqueue).toHaveBeenCalledTimes(1);
+    expect(mockGetInstallationOctokit).toHaveBeenCalledTimes(1);
+  });
+
   it('one suite event yields exactly one job, however many checks it holds', async () => {
     // The fan-out worry resolves by construction: the suite is one event.
     mockGetInstallationOctokit.mockResolvedValue(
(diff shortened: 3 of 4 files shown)


santthosh:
Review's warning addressed — it was correct and it was mine.

`handleCheckSuiteEvent` resolved a client for the ownership check, then `enqueueRereview` resolved another for the same installation: two token exchanges per Re-run, and a second chance to fail *after* ownership had already passed.

Applied the suggested shape: `enqueueRereview` takes an optional already-resolved client and falls back when absent, so the `check_run` path is unchanged. Added a test asserting `getInstallationOctokit` is called exactly once per suite event, so the redundancy cannot come back quietly.

Forced gate green again — build 12/12, typecheck 20/20, test 21/21, `Cached: 0`.

Issue #564, opened 2026-09-06
Phase 1b: reorder prompt segments so all six agents share a cacheable prefix


santthosh:
Split out of #489. **Depends on #489 landing first.**

## Why this is separate

#489's acceptance asked for rendered prompts to be *"byte-identical apart from the deleted date and the deliberate reordering."* Those fight each other: reordering the prompt **is** a behaviour change — the model sees different input and may produce different reviews. "Byte-identical apart from the part that isn't" cannot be asserted, and shipping both together means neither claim holds.

So the split is by **what can be proven**:

| | Scope | Guarantee |
|---|---|---|
| **#489 (1a)** | Segments, providers, ordering lint, floor transform, decorator retype, date deletion | Rendered output differs from today by **exactly the date line** — provable by golden test |
| **this (1b)** | The reordering | **Deliberately** changes prompts; verified by a fixture run and reviewed as a behaviour change |

## What

Today's order is interleaved — static directives (`INTENT_CLAIMS_DIRECTIVE`, `FILE_REQUEST_INSTRUCTION`) sit *after* per-repo content, and each agent's distinct template sits in front of the shared diff, so the six agents share no cacheable prefix at all.

Target:

```
system:    static directives → conventions [break] → PR context + diff [break]
messages:  agent-specific template + agent-mode → fetched files
```

That inversion is what lets **one cache write serve all six agents**, and it matches standard prompting guidance (long context first, specific instruction last).

## Verification — the point of the split

This cannot be verified by asserting byte-equality, because it deliberately breaks it. It needs:

- A golden test updated to the NEW order, so the change is visible in review rather than implicit
- A **fixture run** confirming review quality did not regress — the graded suite is the only instrument that can say whether reordering changed what the model finds
- The cache-hit rate from #490 as the confirming signal that the prefix is genuinely shared

Reordering a prompt to save money and silently degrading review quality would be a bad trade made invisibly. That risk is the whole reason this is its own PR.

## Acceptance

- [ ] Segments are emitted in the target order
- [ ] The ordering lint (from #489) passes on the new order
- [ ] Golden prompt tests updated, with the diff reviewed deliberately
- [ ] A graded fixture run shows no regression in review quality
- [ ] #490 can place a cache breakpoint that serves all six agents

santthosh:
## 🚚 Ship tracker (auto-maintained)

**Scope:** Reorder prompt segments so the six agents share a cacheable prefix.
**Plan:** 1 PR.

[...]

Issue #490, opened 2026-08-26
Phase 2: prompt caching at stability boundaries


santthosh:
Phase 2 of #488. Depends on #489.

## Why

Cache reads run ~10% of input cost, and diff + file contents dominate input tokens in every review. This cuts **production** cost on every review permanently — a larger prize than the e2e bill that started #488.

It also sits second in the sequence deliberately: it **validates Phase 1's stability labels against reality**. A mislabeled segment shows up immediately as a zero cache-hit rate, before the slower replay work starts depending on those labels being right.

### Breakpoints

`cache_control` at stability boundaries, max 4 per request. **This lives in the provider, never in core** — core emits segments, each provider decides:

| Provider | Behaviour |
|---|---|
| Bedrock / Anthropic | explicit `cache_control` breakpoints |
| OpenAI-compatible (litellm) | no cache API — prefix caching is automatic, so Phase 1's stable-first ordering already earns it |
| Ollama | none; renders as today |

For Bedrock this is a small change to `buildAnthropicBody` (`packages/llm-bedrock/src/bedrock-provider.ts:78`) — `content: [{ type: 'text', text, cache_control: {...} }]` instead of a bare string. **Do not adopt the Mantle client** (rationale in #488 → Settled).

### Reorder the verifier — the biggest single win

`verifyFindings` (`packages/core/src/agents/reviewer.ts:2075`) builds:

```
promptHead + finding + "--- Complete current file ---" + content
```

The file content is **last**, so ten findings in one file re-send that whole file ten times at full price. Put the file above the finding and `(head + file)` caches across every finding in that file.

### ⚠️ Cost accounting — this is a BILLING change, not a display change

`estimatedCostUsd` is metered spend, not a label:

```ts
await recordReview(dynamodb, INSTALLATIONS_TABLE, String(installationId),
                   result.estimatedCostUsd, prNumberCommitSha, stripe, ossRepoContext);
// packages/lambda/src/handlers/review-agent.ts:1175 — :1185 logs "[billing] REVENUE LEAK" on failure
```

It drives Stripe charges, OSS grant drawdown, and auto-reload. Getting it wrong here is a revenue bug.

**The failure mode is under-billing.** The API returns `input_tokens` as *uncached input only*; `cache_read_input_tokens` and `cache_creation_input_tokens` are separate fields. The providers read neither:

```ts
[...]

Issue #488, opened 2026-08-26
[Tracker] Prompt structure → caching → cassettes → replayable e2e


santthosh:
Tracker. Full design record: `docs/feat/20260826-01-prompt-structure-caching-replay.design.md`.

## Why this exists

Started as "e2e is expensive, can we replay saved LLM responses?" and ended at a prompt-architecture change, because three findings turned out to share one root cause.

**1. The e2e bill is not somewhere we can wrap it.** The gate opens real PRs in `mergewatch/fixtures` against the deployed dev Lambda, which hardcodes its provider at module scope:

```ts
const llm = new BedrockLLMProvider();   // packages/lambda/src/handlers/review-agent.ts:149
```

A cassette layer only reduces that bill if it ships *inside the deployed artifact* — the one that also serves production. One misconfigured env var and prod reviews come from stale recordings.

**2. We do not cache prompts, and cannot.**

```ts
messages: [{ role: 'user', content: prompt }]   // packages/llm-bedrock/src/bedrock-provider.ts:87
```

No `system` field. Template, conventions, PR context, diff, and fetched file contents are concatenated into one user message, and `cache_control` appears in none of the four providers.

**3. Prompts are not deterministic.** `reviewer.ts:213` puts `Current date: ...` in every agent prompt — invalidates any cache daily, and would stale an entire cassette corpus every midnight.

### The root cause

**`buildPrompt` returns a `string`.** A flat string cannot express which of its parts are stable. Prompt caching needs that to place breakpoints, cassette keys need it to survive unrelated edits, and blast-radius detection needs it to scope re-recording. Three problems, one missing distinction.

## The structure

```ts
type PromptStability =
  | 'static'    // frozen in source: directives, agent templates
  | 'per-repo'  // conventions, tone, custom rules
  | 'per-pr'    // PR context, diff
  | 'per-call'; // fetched files, the finding under verification, previous findings

type PromptSegment = { id: string; text: string; stability: PromptStability };
```

| Need | Falls out as |
|---|---|
| Prompt caching | a breakpoint at each stability boundary |
| Cassette keys | hash of segment ids + content hashes, volatile segments excluded **by declaration** |
| Blast radius | the `static` segments' hash *is* the prompt lockfile |

One ordering discipline, three payoffs. That is the argument for doing it as one change.

## Why the order has to change, not just gain breakpoints

[...]

Issue #543, opened 2026-09-04
A blocking org agent almost never blocks: the admin's severity is a fallback the model can override


santthosh:
## What happens

An org admin creates a custom agent with severity **critical** and enforcement **blocking**. The documented contract (#235) is that a critical finding from that agent forces `REQUEST_CHANGES` plus a failing check titled `Blocked by org agent: <name>`, regardless of the merge score.

It does not. Driven by hand against a deployed stage during the v0.6.1 manual verification, with the agent set to `severity: critical, enforcement: blocking`, the result was:

```
MergeWatch Review: success — 5/5 — 2 findings (no blocking critical)
review state: APPROVED
findings:     🔵 Info (2)
```

The agent ran and produced its findings. The check run's own title says why nothing blocked: **no blocking critical**.

## Why

`packages/core/src/agents/reviewer.ts:1237`:

```ts
severity: f.severity || agentDef.severityDefault,
```

`severityDefault` applies only when the model omits a severity. The model returned `info` for both findings, so `info` won over the admin's configured `critical`. Blocking fires only on `critical`, so the enforcement never engages.

The net effect: **an admin cannot make a blocking agent block.** They can set `critical`, and the model overrides it per finding. Whether the policy is enforced depends on how the model happened to rate that instance — which is the one thing the admin configured the field to control.

## The question this needs answered first

The field is named `severityDefault` and the dashboard labels it "default severity", which argues the current behaviour is intentional — a default the model may refine.

But `enforcement: blocking` only means something if severity is predictable. As built, the two settings contradict each other: one says "this is policy", the other says "unless the model disagrees". Either

- **severity is a floor/override for the agent that declared it** (a `critical` agent's findings are critical), which makes blocking usable and matches what an admin setting it expects; or
- **severity stays a default**, and then blocking needs a different trigger — e.g. any finding from a blocking agent blocks, with severity only affecting presentation.

The first is the smaller change and matches the #235 write-up. Worth an explicit decision either way, because today's behaviour is neither.

## Related: the fixture cannot detect this

`68-org-custom-agents` states its failure tell as:

> ❌ The repo `.mergewatch.yml` agent shadows or disables the org agent of the same name (precedence inverted — the tell is an `info` TODO finding instead of a `critical` one).

[...]

PR #565, opened 2026-09-06
feat(core): prompts become ordered segments with stability labels (#489)

### LINKED ISSUE #489: Phase 1a: buildPrompt returns PromptSegment[] — no reordering, byte-identical but for the date
Phase 1 of #488. **Nothing else in that tracker can start until this lands** — cassette keys derive from segment structure, so building anything on flat-string prompt hashes invalidates it the day this ships.

## Why

`buildPrompt` (`packages/core/src/agents/reviewer.ts:194`) returns a `string`. A flat string cannot express which of its parts are stable — which is what prompt caching needs to place breakpoints, what cassette keys need to survive unrelated edits, and what blast-radius detection needs to scope re-recording.

## What

```ts
type PromptStability =
  | 'static'    // frozen in source: directives, agent templates
  | 'per-repo'  // conventions, tone, custom rules
  | 'per-pr'    // PR context, diff
  | 'per-call'; // fetched files, the finding under verification, previous findings

type PromptSegment = { id: string; text: string; stability: PromptStability };
```

- `buildPrompt` and the verifier / orchestrator prompt builders return `PromptSegment[]`
- `ILLMProvider.invoke` / `invokeStructured` take segments; each provider renders them. A provider that cannot cache renders them into the same string it builds today
- **Ordering lint:** segments must be non-decreasing in volatility. A `static` segment after a `per-pr` one fails the build
- Delete the `Current date:` line (`reviewer.ts:213`) — see #488 open question 3

### Target order

Today is interleaved: static directives (`INTENT_CLAIMS_DIRECTIVE`, `FILE_REQUEST_INSTRUCTION`) sit *after* per-repo content, and the volatile date sits *before* the diff.

```
system:    static directives → conventions [break] → PR context + diff [break]
messages:  agent-specific template + agent-mode → fetched files
```

The inversion is what later lets one cache write serve all six agents — today each agent's distinct template sits in front of the shared diff, so they share no cacheable prefix. It also matches standard prompting guidance (long context first, specific instruction last).

## Two traps

**`applyConfidenceFloor` must become a segment transform, not a regex over the assembled prompt.** It currently rewrites the whole prompt string in the `floorLlm` decorator. Left as-is, any repo setting a non-default `minConfidence` silently breaks prefix stability — a cache miss *and* a cassette miss, with no visible cause.

**The decorator chain moves with it.** `runReviewPipeline` wraps `invoke` four deep — tracking → capped → confidence-floor → truncation-retry — all typed against `prompt: string`.

## Acceptance criteria

- [ ] `buildPrompt`, the verifier prompt, and the orchestrator prompt return `PromptSegment[]`
- [ ] All four providers render segments; ollama/litellm output is byte-identical to today
- [ ] Ordering lint fails the build on a volatility inversion, with a test proving it
- [ ] `applyConfidenceFloor` operates on a segment; a non-default `minConfidence` leaves all other segments byte-identical
- [ ] The four-layer decorator chain compiles and behaves unchanged
- [ ] **No behaviour change:** rendered prompts are byte-identical apart from the deleted date and the deliberate reordering, and a test asserts it
- [ ] No changes in `packages/lambda` or `packages/server` — this is core + providers only

santthosh:
Phase 1a of #488. Closes #489.

## Why

`buildPrompt` returned a flat `string`. A flat string cannot express which of its parts are stable — which is what prompt caching needs to place breakpoints, what cassette keys need to survive unrelated edits, and what blast-radius detection needs to scope re-recording. One missing distinction, three blocked payoffs.

**Nothing else in #488 can start until this lands** — cassette keys derive from segment structure.

## What changed

- `buildPrompt`, the agentic fetcher and `applyConfidenceFloor` speak `PromptSegment[]`
- `ILLMProvider` accepts `string | PromptSegment[]`, so the three migrated builders convert **without touching all 26 call sites** — an unmigrated caller keeps passing a string and behaves exactly as before
- All four providers render via `renderPrompt`
- New ordering lint: segments must be non-decreasing in volatility
- `Current date:` deleted (#488 open question 3, decided)

## Segments are deliberately coarse — and that is the honest choice

`head` is labelled `per-pr` because the agent-mode block inside it varies per PR, even though the template and intent-claims directive are static.

Splitting them apart would put a `static` segment after a `per-pr` one — the exact volatility inversion the new lint rejects, and what **#564** fixes by reordering. Until that lands, merging and labelling honestly is right: claiming a stability the text does not have produces a cache prefix that never hits, with nothing to say why.

So this phase delivers the *type and the plumbing*, not the caching win. That arrives with #564 + #490.

## The trap the issue predicted, confirmed

`applyConfidenceFloor` was a regex over the whole assembled prompt. Left that way, any repo setting a non-default `minConfidence` rewrites text anywhere in the prompt — a cache **and** cassette miss for that repo, with no visible cause. It now rewrites per segment; a test asserts every other segment is byte-identical and object-identical.

## A bug this nearly shipped

`bedrock`, `litellm` and `ollama` passed `prompt` straight into `content:` in an **untyped JSON body**. Widening the parameter gave TypeScript nothing to check — so `PromptSegment[]` would have been serialised as a JSON **array** into every request, breaking every review, and compiling perfectly cleanly.

Every render site is now explicit, and `ollama-provider.test.ts` asserts the wire carries a string. **Verified it catches the bug** by reverting the render and re-running: the test fails.

Same shape as the four configuration failures this week — it compiles, it looks right, and it is silently wrong.

## Tests

- `prompt-segment.test.ts` — 9: rendering is pure concatenation (a separator would reflow every prompt), inversion detection both ways, error names both segments, `mostVolatile` for honest merged labels
- `prompt-golden.test.ts` — 9: **there were no prompt snapshot tests before this.** Asserts the rendered shape, that the date line is gone, that the PR context it sat above survives, non-decreasing volatility, honest labels, and stable segment ids
- `ollama-provider.test.ts` — 2 wire-format tests

## Scope

No changes in `packages/lambda` or `packages/server`. **No reordering** — that is #564, which deliberately breaks byte-equality and is verified by a graded fixture run rather than by assertion.

## Limit

The golden tests pin the rendered *shape* and the date deletion, not a byte-for-byte capture of every agent's real prompt. A full corpus snapshot would be stronger; this covers the structural claims the refactor actually makes. The E2E gate is the backstop — every fixture exercises real prompts end to end.

### DIFF
--- a/packages/core/src/agents/prompt-golden.test.ts
+++ b/packages/core/src/agents/prompt-golden.test.ts
@@ -1,0 +1,111 @@
+import { describe, it, expect } from 'vitest';
+import { buildPrompt } from './reviewer.js';
+import { renderPrompt, findVolatilityInversion } from '../llm/prompt-segment.js';
+import { applyConfidenceFloor } from './prompts.js';
+
+/**
+ * #489 — the guarantee that makes this refactor safe.
+ *
+ * `buildPrompt` went from returning a string to returning segments, and the
+ * provider interface widened to accept either. The claim is that rendered
+ * output is byte-identical to before, EXCEPT the deliberately deleted
+ * `Current date:` line (#488 open question 3).
+ *
+ * There were no prompt snapshot tests before this. Without them the claim
+ * would be an assertion rather than a fact, which is exactly the kind of
+ * unverifiable "no behaviour change" this repo has been bitten by.
+ *
+ * Reordering is NOT in this phase — see #564. That change deliberately breaks
+ * byte-equality and is verified by a graded fixture run instead.
+ */
+const CONTEXT = {
+  owner: 'octo',
+  repo: 'demo',
+  prNumber: 7,
+  prTitle: 'Add retry policy',
+  prBody: 'Adds a shared retry helper.',
+} as const;
+
+const TEMPLATE = [
+  'AGENT TEMPLATE',
+  'TONE_PLACEHOLDER',
+  'CONVENTIONS_PLACEHOLDER',
+  'AGENT_MODE_PLACEHOLDER',
+  'FILE_REQUEST_PLACEHOLDER',
+].join('\n');
+
+const build = (over: Partial<{ agentic: boolean; conventions: string; agentAuthored: boolean }> = {}) =>
+  buildPrompt(
+    TEMPLATE, 'DIFF BODY', CONTEXT as never,
+    over.agentic ?? false, undefined, over.conventions, over.agentAuthored ?? false,
+  );
+
+describe('#489 — buildPrompt returns segments', () => {
+  it('renders to the exact shape the flat string produced', () => {
+    const rendered = renderPrompt(build());
+    // The assembled order is unchanged: template, intent-claims, PR context, diff.
+    expect(rendered).toContain('AGENT TEMPLATE');
+    expect(rendered.indexOf('--- PR Context ---')).toBeGreaterThan(rendered.indexOf('AGENT TEMPLATE'));
+    expect(rendered.indexOf('--- Diff ---')).toBeGreaterThan(rendered.indexOf('--- PR Context ---'));
+    expect(rendered.endsWith('DIFF BODY')).toBe(true);
+  });
+
+  it('has NO date line — the one deliberate difference from before', () => {
+    // It invalidated any prompt cache daily and would invalidate a cassette
+    // corpus every midnight, for context the review does not use.
+    expect(renderPrompt(build())).not.toMatch(/Current date:/);
+  });
+
+  it('still carries the PR context the date line used to sit above', () => {
+    // Guard against deleting more than intended.
+    const r = renderPrompt(build());
+    expect(r).toContain('Repository: octo/demo');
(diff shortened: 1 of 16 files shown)


santthosh:
All five info findings are the same class and I'm justifying rather than changing them.

Each flags a comment that references deferred work by **tracked issue number** — #564 (the reorder), #490 (cache breakpoints), #488 Q3 (the date decision). That is the opposite of a TODO: a TODO rots because nothing tracks it, whereas `// see #564` is discoverable from both ends and dies when the issue closes. Removing them would leave the coarse-segment choice and the un-placed breakpoints looking like oversights rather than sequenced work.

One is slightly off on its facts: `reviewer.ts:219` is described as *"referencing unresolved open question (#488 Q3)"*, but the comment says **"(#488 open question 3, decided: delete.)"** — the question is resolved and the comment records the decision, which is why the line was removed.

Worth noting these come from the `no-todo` org custom agent (advisory), which is doing exactly what it is configured to do. Its severity is model-chosen rather than the configured `critical`, which is #543.

PR #567, opened 2026-09-06
feat(core): cache-aware token accounting and pricing (#490, phase A)

santthosh:
Phase A of #490. Part of #488.

## Why this lands before any breakpoints

`estimatedCostUsd` is **metered spend**, not a label — it drives Stripe charges, OSS grant drawdown and auto-reload. The Anthropic/Bedrock APIs report `input_tokens` as **uncached input only**, with `cache_read_input_tokens` and `cache_creation_input_tokens` as separate fields that no provider read:

```ts
{ inputTokens: parsed.usage.input_tokens ?? 0, outputTokens: parsed.usage.output_tokens ?? 0 }
```

Place breakpoints first and cached tokens vanish from the count entirely. True cost drops with caching; **reported cost would drop further**, charging zero for tokens billed at 0.1× and 1.25×. That is under-billing, arriving through a door the `REVENUE LEAK` alarm does not watch.

So the meter is fixed first, while nothing is using the cache.

## What changed

- `TokenUsage` gains `cacheReadInputTokens` / `cacheWriteInputTokens`
- Bedrock and Anthropic populate them at **all four** extraction sites
- `TokenAccumulator` tracks them per model, reported separately from uncached input
- `estimateCost` prices them

## Rates are multipliers, not pricing fields

```ts
export const CACHE_READ_MULTIPLIER = 0.1;
export const CACHE_WRITE_MULTIPLIER = 1.25;
```

Adding `cacheReadPer1M` / `cacheWritePer1M` would mean editing ~40 rows **and breaking every `customPricing` override**: a self-hosted `.mergewatch.yml` `pricing:` block carries two keys today, so any override not updated would silently price cache traffic at **zero**. Deriving from `inputPer1M` keeps existing two-key overrides correct by construction — and there is a test that pins exactly that.

## No behaviour change

Nothing caches yet, so every new count is zero and cost is arithmetically identical. litellm and ollama have no cache API and omit the fields, which reads as zero — asserted.

## Tests

11 added. The ones that matter are the billing ones:

- reads and writes priced at their own rates, not zero
- **a cached review costs more than counting uncached input alone** — the under-billing failure stated directly (0.30 vs 0.60 on the same review)
- an existing **two-key** `customPricing` override still prices cache traffic
- absent cache fields are arithmetically identical to before
- the Anthropic provider **extracts** the fields — verified to fail without the change by reverting the provider

**Worth noting:** vitest passed 9/9 on the provider tests while `typecheck` failed on an invalid cast in one of them. Vitest does not typecheck; the two are run separately for exactly this reason.

## Not in this PR

Breakpoints and the verifier reorder are Phase B. Also **not merging either until #566's E2E gate is green** — Phase B caches the prefix that PR created, and if the reorder regressed review quality it needs reverting rather than caching.

## Acceptance criterion I cannot honestly meet

> a test asserts `cache_read_input_tokens > 0` on agents 2..6

A unit test cannot establish this — mocked, it proves only that I wrote the number into the mock. It needs a real API round-trip. What is provable is that the request carries `cache_control` (Phase B) and that whatever comes back is priced correctly (this PR). Whether Bedrock honours it is measurable from the gate's `Suite cost` line across runs, and I will report it that way rather than claim the criterion met.

### DIFF
--- a/packages/core/src/llm/cache-accounting.test.ts
+++ b/packages/core/src/llm/cache-accounting.test.ts
@@ -31,7 +31,8 @@ describe('estimateCost with cache traffic', () => {
 
   it('a cached review costs MORE than counting uncached input alone', () => {
     // The under-billing failure stated directly: if cache traffic were ignored,
-    // this review would be billed 0.30 when it actually cost 0.63.
+    // this review would be billed 0.30 when it actually cost 0.60 —
+    // 100k uncached at $3/1M = $0.30, plus 1M cache-read at $3/1M x 0.1 = $0.30.
     const ignored = estimateCost(MODEL, 100_000, 0, PRICING)!;
     const actual = estimateCost(MODEL, 100_000, 0, PRICING, { readTokens: 1_000_000, writeTokens: 0 })!;
     expect(ignored).toBeCloseTo(0.3, 10);
--- a/packages/core/src/llm/cache-accounting.test.ts
+++ b/packages/core/src/llm/cache-accounting.test.ts
@@ -1,0 +1,90 @@
+import { describe, it, expect } from 'vitest';
+import { estimateCost, CACHE_READ_MULTIPLIER, CACHE_WRITE_MULTIPLIER } from './pricing.js';
+import { TokenAccumulator } from './token-accumulator.js';
+
+/**
+ * #490 Phase A — cache-aware accounting.
+ *
+ * `estimatedCostUsd` is metered spend, not a label: it drives Stripe charges,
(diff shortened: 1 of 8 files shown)


santthosh:
Fixed — the review was right and it was a real error.

$0.30 uncached + $0.30 cache-read = **$0.60**; the assertion was correct and my comment said 0.63. A wrong number in a comment explaining billing arithmetic is exactly where it misleads someone auditing the meter, so this was worth the catch rather than a nit. Also spelled out the working in the comment so the next reader can check it without recomputing.

PR #568, opened 2026-09-06
feat(core): cache_control breakpoints and verifier reorder (#490, phase B)

### LINKED ISSUE #490: Phase 2: prompt caching at stability boundaries

santthosh:
Phase B of #490. Closes #490. Builds on #567 (accounting) and #566 (shared prefix).

## The verifier reorder — the biggest single win

`verifyFindings` built `head + finding + FILE`, with the file **last**. Ten findings in one file re-sent that whole file **ten times at full price**.

The file now sits above the finding, so `head + file` is an identical prefix for every finding in that file and is served from cache after the first. Its segment is `per-pr` — the file is fixed at the PR head for the whole review — while only the finding is `per-call`. That gives `static → per-pr → per-call`, and the breakpoint falls exactly where it should.

## Breakpoints

`toCacheableBlocks` emits Anthropic content blocks with `cache_control` at stability boundaries. Two decisions worth stating:

**It groups consecutive same-stability segments first.** A breakpoint inside a run of equally-stable text buys nothing and spends one of the four the API allows.

**It never marks the last group.** Caching a prefix that includes the most volatile segment writes an entry nothing can ever read — paying the 1.25× write premium for zero reads. That is a way to make caching cost *more*, so it is prevented by construction and asserted by a test.

Providers switch to block form **only** when a prompt genuinely has more than one stability level. A string, or segments that are all one stability, takes the old path and is byte-identical on the wire — so unmigrated callers cannot start paying a write premium by accident.

## Unaffected

litellm and ollama have no cache API. Their prefix caching is automatic, which #564's stable-first ordering already earns.

## Tests

9 added. The first is the one that matters most:

- **the wire content equals the rendered prompt** — caching must not change what the model reads; if that ever fails, cost is the least of the problems
- consecutive same-stability segments collapse into one block
- every boundary marked except the last
- the 4-breakpoint API limit is never exceeded
- a plain string passes through as a single **unmarked** block
- the verifier shape: the cached prefix is identical across findings, contains the file, and does **not** contain the finding

## What this does not prove — and how it will be judged

**That Bedrock actually honours the breakpoints.** #490 asks for a test asserting `cache_read_input_tokens > 0` on agents 2..6; mocked, that proves only that I wrote the number into the mock. It needs a real round-trip.

What is provable here is that the **request** carries `cache_control` at the right boundaries, and (from #567) that whatever comes back is **priced correctly** — including cache reads at 0.1× and writes at 1.25×, so a working cache cannot silently under-bill.

The real evidence is the gate's `Suite cost` line across runs. Three baselines: **$4.45 → $4.48 → $4.55**, all with 47/48 graded clean. If caching works, this run should fall below them; if it does not, cost stays flat and I will say so rather than claim the win.

This also **deliberately changes the verifier prompt**, so review quality is again only judgeable by the graded suite.

### DIFF
--- a/packages/core/src/agents/reviewer.ts
+++ b/packages/core/src/agents/reviewer.ts
@@ -2334,18 +2334,30 @@ export async function verifyFindings(
         return { keep: false };
       }
 
-      const prompt = `${verifierPromptHead}
-
---- Finding ---
-File: ${f.file}
-Line: ${f.line}
-Severity: ${f.severity}
-Title: ${f.title}
-Description: ${f.description}
-Suggestion: ${f.suggestion}
-
---- Complete current file: ${f.file} ---
-${content}`;
+      // #490 — the FILE goes ABOVE the finding, and that ordering is the
+      // single biggest saving in the pipeline. This used to send
+      // `head + finding + file`, so ten findings in one file re-sent that
+      // whole file ten times at full price. With the file above, the prefix
+      // `head + file` is identical for every finding in that file and is
+      // served from cache after the first.
+      //
+      // Labels: the file content is `per-pr` because it is fixed at the PR
+      // head for the whole review; only the finding differs per call. That
+      // gives static → per-pr → per-call, which is where the breakpoint goes.
+      const prompt: PromptSegment[] = [
+        { id: 'verifier-head', stability: 'static', text: verifierPromptHead },
+        {
+          id: 'verifier-file',
+          stability: 'per-pr',
+          text: `\n\n--- Complete current file: ${f.file} ---\n${content}`,
+        },
+        {
+          id: 'verifier-finding',
+          stability: 'per-call',
+          text: `\n\n--- Finding ---\nFile: ${f.file}\nLine: ${f.line}\nSeverity: ${f.severity}\n`
+            + `Title: ${f.title}\nDescription: ${f.description}\nSuggestion: ${f.suggestion}`,
+        },
+      ];
 
       try {
         // #390 — prefer schema-constrained verdicts; unsupported providers
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -2,6 +2,8 @@
 export type { ILLMProvider, TokenUsage, LLMInvokeResult, LLMSamplingConfig, LLMStructuredResult } from './llm/types.js';
 export type { PromptSegment, PromptStability, PromptInput } from './llm/prompt-segment.js';
 export { renderPrompt, findVolatilityInversion, assertNonDecreasingVolatility, mostVolatile } from './llm/prompt-segment.js';
+export { toCacheableBlocks, hasCacheableBoundary, MAX_CACHE_BREAKPOINTS } from './llm/cache-blocks.js';
+export type { AnthropicContentBlock } from './llm/cache-blocks.js';
 export { normalizeLLMResult, StructuredOutputUnsupportedError } from './llm/types.js';
 export { TokenAccumulator, TrackingLLMProvider } from './llm/token-accumulator.js';
 export { estimateCost, DEFAULT_PRICING, parseEnvModelPricing } from './llm/pricing.js';
--- a/packages/core/src/llm/cache-blocks.test.ts
+++ b/packages/core/src/llm/cache-blocks.test.ts
@@ -1,0 +1,90 @@
+import { describe, it, expect } from 'vitest';
+import { toCacheableBlocks, hasCacheableBoundary, MAX_CACHE_BREAKPOINTS } from './cache-blocks.js';
+import { renderPrompt, type PromptSegment } from './prompt-segment.js';
+
+const seg = (id: string, stability: PromptSegment['stability'], text: string): PromptSegment =>
+  ({ id, stability, text });
+
+const REVIEW = [
+  seg('shared', 'static', 'DIRECTIVES'),
+  seg('conv', 'per-repo', '\n\nCONVENTIONS'),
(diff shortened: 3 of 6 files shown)


santthosh:
Rebased onto `main` — it now carries #571 (`maxFindings` enforced) and Phase A (`a9a6c111`, cache-aware billing), both of which reached production on [run 34151044399](https://github.com/mergewatch/mergewatch.ai/actions/runs/34151044399) with the gate green at **47/48**.

That matters for this PR specifically: Phase A had to be live *before* anything starts using the cache, or cached tokens would vanish from the bill and reported cost would fall further than true cost. That precondition is now met.

Full gate re-run against the rebased tree: build 12/12, typecheck 20/20, test 21/21, all `Cached: 0`.

**The prediction this PR is judged on:** four consecutive baselines now stand at **$4.45 → $4.48 → $4.55 → $4.53**, all at 47/48. If Bedrock honours the breakpoints, the next `Suite cost` should fall below that band for the first time. If it stays flat, caching is not working and I will report that rather than claim the win.

Issue #561, opened 2026-09-04
Nothing pins the cost parser to the cost formatter — a format change would silently zero the suite figure


santthosh:
Follow-up to #536, raised by the automated review on mergewatch/fixtures#1859.

## The coupling

`grade-run.mjs` (fixtures repo) parses the cost block that `comment-formatter.ts` (this repo) emits:

```
| **Tokens** | 115,256 in · 2,551 out · 117,807 total |
| **Est. cost** | ~$0.2424 (LLM only) |
| **Est. cost** | ~$0.1234 this run · ~$0.5678 total for PR (LLM only) |
```

Two repos, no shared type, no build-time check. If the formatter's details table changes shape, the parser stops matching and every fixture reports **unknown**.

## Why it is worth tracking rather than ignoring

The failure is *visible* — `N fixture(s) reported no cost` is printed and the total shrinks — so this is not the silent-$0 hole #536 closed. But it is quiet: a suite that suddenly reports `~$0.00 across 0 reviewed fixtures` reads like a cheap run rather than a broken parser, and the number would be wrong in the direction that makes the cost work look finished.

That is the same shape as the four configuration failures this week (#544, #545, #558, #559): the code is right, an out-of-band coupling breaks, and nothing says so.

## Options

1. **Assert the pairing in the formatter's tests.** Cheapest: a test in this repo that renders a review comment and runs the same regexes the grader uses, failing if they stop matching. Duplicates the patterns but catches the break at the right moment — when the formatter changes.
2. **Emit a machine-readable block.** An HTML comment carrying `{"costUsd":…,"inputTokens":…}` alongside the human table. Parsing stops depending on prose formatting entirely. More robust; adds a payload to every review comment.
3. **Fail loud on total absence.** If *every* graded fixture reports unknown, that is a parser break, not a cheap run — say so explicitly rather than printing `$0.00`.

Option 3 is worth doing regardless and is a few lines. Option 1 or 2 decides how the coupling is held.

## Acceptance

- [ ] A formatter change that breaks the cost block fails a test, rather than surfacing as a quiet zero
- [ ] A run where every fixture reports unknown is reported as a parser failure, not as a cheap run
- [ ] A decision recorded on whether the comment carries a machine-readable cost payload

santthosh:
## 🚚 Ship tracker (auto-maintained)

**Scope:** Stop a formatter change from silently zeroing the suite cost figure.
[...]

PR #574, opened 2026-09-08
feat(core): emit a machine-readable cost payload (#561, phase 2)

santthosh:
Phase 2 of 3 for #561. Phase 1 is fixtures#2241; Phase 3 is the grader change that reads this.

## Why a payload rather than a test

#561 offered three options. **Option 1 — assert the pairing in a test here — has the flaw the issue is about.** The parser lives in a *different repo*, so a test in this one cannot import it; it would have to copy the regexes, and a copied regex drifts from the real parser silently. That is exactly the failure mode being fixed.

Compare `review-findings.test.ts`, the existing drift test: it works because it imports the *real* core function. Not available across a repo boundary.

**Option 2 removes the coupling instead of testing it.**

```
<!-- mw-cost:{"inputTokens":115256,"outputTokens":2551,"estimatedCostUsd":0.2424} -->
```

Invisible when rendered, and the pattern is already idiomatic here — `<!-- mw-fp:… -->` at `client.ts:719` does the same for inline fingerprints.

## Three details

**Both cost figures are separate fields.** The prose table switches sentence form once a PR is reviewed twice (`~$X this run · ~$Y total for PR`), which the regex has to special-case. The payload keeps `estimatedCostUsd` and `cumulativeCostUsd` distinct, so a consumer never has to know which form it is looking at.

**Emitted even when empty.** `{}` says *"this deployment emits payloads and this review genuinely had no cost"*; no payload at all says *"I could not tell"*. Those are different, and the ambiguity is precisely what #561 is about.

**`parseCostPayload` returns null rather than throwing.** Throwing would take down a whole grading step rather than one note — the same failure shape as the manifest-escaping bug caught on fixtures#2122. Null is also what lets a consumer fall back to the prose.

## Tests

10 added. The two that matter most assert **the payload agrees with the rendered table** — two representations of one number drift the first time someone edits one, and the grader will *prefer* the payload, so a disagreement would bill against numbers nobody can see.

## Rollout

Phase 3 must land **after** this reaches production, and must **keep the prose fallback** until every deployment emits the payload — otherwise a review from an older stage reports unknown cost. The two repos cannot land atomically; the fallback is what makes the overlap safe.

Between the two, the payload exists and nothing reads it. Harmless.

### DIFF
--- a/packages/core/src/comment-formatter.test.ts
+++ b/packages/core/src/comment-formatter.test.ts
@@ -188,6 +188,46 @@ describe('formatReviewComment', () => {
     expect(result).not.toContain('```mermaid');
   });
 
+  // #561 — the payload and the prose must describe the SAME review.
+  //
+  // Two representations of one number will drift the first time someone edits
+  // one of them. The grader will prefer the payload, so a payload that
+  // disagreed with the visible table would silently bill against numbers
+  // nobody could see.
+  it('#561 — the machine-readable payload agrees with the rendered table', async () => {
+    const { parseCostPayload } = await import('./comment-formatter.js');
+    const result = formatReviewComment(baseOptions({
+      inputTokens: 115256,
+      outputTokens: 2551,
+      estimatedCostUsd: 0.2424,
+    }));
+
+    const payload = parseCostPayload(result);
+    expect(payload).not.toBeNull();
+    expect(payload!.estimatedCostUsd).toBe(0.2424);
+    expect(payload!.inputTokens).toBe(115256);
+    expect(payload!.outputTokens).toBe(2551);
+
+    // …and those exact figures are what the human table shows.
+    expect(result).toContain('$0.2424');
+    expect(result).toContain('115,256 in');
+    expect(result).toContain('2,551 out');
+  });
+
+  it('#561 — a re-review payload carries both figures the prose splits', async () => {
+    const { parseCostPayload } = await import('./comment-formatter.js');
+    const result = formatReviewComment(baseOptions({
+      inputTokens: 10, outputTokens: 5,
+      estimatedCostUsd: 0.1, cumulativeCostUsd: 0.75,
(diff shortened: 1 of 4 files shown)


santthosh:
## This is merged but **not gate-verified** — recording that rather than leaving it implicit

Its own deploy run [34250657048](https://github.com/mergewatch/mergewatch.ai/actions/runs/34250657048) **failed**, and not on a graded assertion:

```
##[error]Timed out after 900s waiting for reviews.
  ✗ 06-docs-only #2246: absent
```

Grading never ran, so no fixture graded this change.

The next run on `main` ([34254674336](https://github.com/mergewatch/mergewatch.ai/actions/runs/34254674336), #577) shows `E2E gate (dev): success` and **contains this commit** — but it selected **0 fixtures**, because #577 was README-only. That is the "0 selected" outcome, not a graded pass. It exercised none of this code, and reporting the two as equivalent would rebuild the exact hole the gate exists to close.

So: **the cost payload is in production, verified by nothing.**

## Why the failure is an implausible cause, and why that is not proof

`06-docs-only` asserts a docs-only PR is **skipped**. The skip path posts a check run and never formats a review comment, so `renderCostPayload` never executes on it. A 900-second timeout with a check run simply absent reads as environmental rather than as anything this diff did.

But implausible is not ruled out, and no graded run has covered it.

## Covered by #578

Not re-running this deploy to chase it. #578 touches the workflows and `release.sh`, so it will select a full suite and grade this payload as a side effect — in a run that is needed anyway.

Two things to check there:
- the `Suite cost` line still parses (~$3.0x) — if the payload broke the **prose** parser this PR was written to protect, the #561 guard from fixtures#2241 will now say `Suite cost: UNKNOWN` instead of a silent `$0.00`
- whether `06-docs-only` times out again. Once is environmental; twice is a pattern worth chasing.

Issue #575, opened 2026-09-08
[Tracker] Release-process gaps found while cutting v0.6.0 and v0.6.1


santthosh:
Cutting v0.6.0 and v0.6.1 exposed a series of gaps in the release process itself. Three were fixed in flight; four are still open and are collected here so the milestone reflects the whole set rather than the leftovers.

## Fixed in flight (no issue existed — the commit is the record)

| Gap | Symptom | Fix |
|---|---|---|
| Release notes were built **after** the tag push | v0.6.0's job died mid-step, leaving a tag with no release — which the "refuse to re-cut an existing tag" guard then blocked from being retried | notes moved to the near side of the irreversible push |
| `grep -oE '\`+'` exits 1 on no match | Under `set -e`, the fence-sizing guard aborted the release in every case *except* the rare one it was written for | `{ … \|\| true; }` |
| `GITHUB_TOKEN` releases emit no `release: published` event | v0.6.0 published **no images** and reported success | the gate dispatches `docker-publish` explicitly and waits |
| `permissions: contents: write` only | v0.6.1 tagged and released, then failed `HTTP 403` dispatching docker-publish — **the release existed with no images**, published by hand afterwards | `actions: write` added (#559) |

That last one is worth keeping visible: it was caught **only** because #513/#514 had already added a guard that refused to call a release successful when no images shipped. Without it, a green run would have hidden a broken release.

## Still open — the four in this milestone

- **#549** — the gate never bumps versions or updates `CHANGELOG.md`. **v0.6.0 shipped with every `package.json` still on `0.5.0`**, and there is no `0.6.0` changelog section. `scripts/release.sh` already does this work and is orphaned: #505 replaced the manual flow and never adopted it.
- **#550** — release notes say *how it was tested*, never *what changed*. v0.6.1's notes omit all eight fixes, including two that made features silently non-functional.
- **#551** — the README documents the **pre-#505 manual flow**, contradicts itself on Docker publishing within the same page, and claims prod needs manual approval (untrue since #428). A self-hoster following it builds an App whose Re-run button is dead.
- **#552** — the whole process is documented only in comments inside `release-gate.yml`. `grep -rl "release-gate" docs/ README.md` returns nothing.

## Sequencing

**#549 first.** #550 needs the same commit-range generator, and #552 should describe the finished process rather than one mid-change. #551 is independent and can go any time.

[...]

Issue #552, opened 2026-09-04
The release gate is undocumented outside its own workflow file


santthosh:
## The gap

The only description of how MergeWatch releases work lives in comments inside `.github/workflows/release-gate.yml`. Those comments are good — they explain why `suite` holds the `e2e-fixtures` lock, why `verify` deliberately holds none (#428's 9.5-hour cancellation), and why notes are built before the tag is pushed (what killed v0.6.0). But nobody reads a workflow file to answer "how do I cut a release", and `grep -rl "release-gate" docs/ README.md` returns nothing.

The practical result: releasing depends on knowledge that is not written down anywhere a person would look, and the README describes a different process entirely (#551).

## Proposed — `docs/releasing.md`

Cover what someone cutting a release actually needs:

- **How to cut one**: dispatch Release Gate with `version` and `candidate_ref`; what `dry_run` does (grades and stops — note it is *not* a cheaper path, it runs the same suite)
- **The three phases and why they are separate jobs**: `suite` grades and holds the fixtures lock; `verify` waits for a human and holds no lock; `release` tags, releases and publishes
- **What the manual verification actually involves** — 32 manual fixtures, and that a green automated suite is not evidence for them
- **The four gate outcomes** (success / failure / skipped / 0 selected) and that they must never be reported interchangeably
- **What happens on failure**: the release is not cut, and re-cutting the same version is blocked by the existing-tag guard until resolved
- **The known cost**: a full graded suite per cut, and that it re-grades a SHA the deploy gate may have graded minutes earlier (#537)
- **Prod deploys are not part of the release** — `main` reaches production on a timer at merge (#428), independent of tagging. Worth stating plainly, because the two are easy to conflate.

## Why it is worth writing rather than assumed

Every failure mode this process has hit was recoverable only by someone who had read the workflow: the tag-then-fail that stranded v0.6.0, the `GITHUB_TOKEN` release that published no images, and the baseline drift that rejected every fixture push. Each is a paragraph in a doc and an afternoon without one.

## Acceptance

- [ ] `docs/releasing.md` exists and covers the above
- [ ] README links to it from the release section (#551)
- [ ] The four gate outcomes are stated with the rule that they are not interchangeable
- [ ] It says plainly that prod deploys happen on merge, not on release

santthosh:
## 🚚 Ship tracker (auto-maintained)

**Scope:** Document the release process outside `release-gate.yml`.
**Plan:** 1 PR.

[...]

Issue #550, opened 2026-09-04
Release notes say how it was tested, never what changed


santthosh:
## What a release note looks like today

The gate's `release` job composes notes from three things — the ref, the graded suite output, and the approver:

```
Released from `main` (commit `d065b724`) via the release gate (#505).

Graded fixture suite:

​```
47 passed · 0 failed · 0 ungraded · 1 skipped · 0 errored
​```

Manual fixtures verified and approved by @santthosh.
```

That is a **provenance record**, and a good one — it is exactly the evidence the gate exists to produce. But it tells a reader nothing about what they are upgrading to. Someone deciding whether to pull `v0.6.1` learns how thoroughly it was tested and nothing about what it fixes.

For v0.6.1 specifically, the notes would omit that it carries eight fixes including two that made features silently non-functional (#526, #544).

## Proposed

Keep the provenance block — it is the differentiator and should stay prominent — and add a **what changed** section above it.

- Generate from conventional commits between the previous tag and the graded SHA, grouped `feat` / `fix` / other. `scripts/release.sh` already implements exactly this for `CHANGELOG.md` (step 4) and is currently orphaned — see #549. One generator should feed both, so the release note and the changelog cannot disagree.
- Link the issues each commit closes; the repo's commit convention already carries them (`fix(core): … (#544) (#547)`).
- Keep it honest about what the suite covered: the existing `NOT VERIFIED: N correctness fixture(s) outside this run` line belongs in the notes too, not just the job log. A release note that shows a green grade without showing what was not run recreates the coverage illusion the gate was built to remove.

## Sequencing

Depends on #549 — both need the same commit-range generator, and building it twice guarantees drift.

## Acceptance

- [ ] Release notes list what changed, grouped by type, with issue links
- [ ] The same generator produces the `CHANGELOG.md` section
- [ ] The provenance block (graded SHA, suite grade, approver) is retained
- [ ] The "not verified" count appears in the notes, not only the log
- [ ] A release with no conventional commits in range produces a sensible note rather than an empty section

PR #578, opened 2026-09-08
feat(release): the gate bumps versions and the changelog (#549)

### LINKED ISSUE #549: The release gate never bumps versions or updates the CHANGELOG — v0.6.0 shipped as 0.5.0
## What is wrong

`v0.6.0` was released on 2026-08-27. Today, with `v0.6.1` being cut:

```
$ gh release list
v0.6.0   Latest   2026-08-27
v0.5.0            2026-08-20

$ grep '"version"' package.json
  "version": "0.5.0",
```

Every workspace `package.json` still says `0.5.0`. `CHANGELOG.md`'s newest entry is `## [0.5.0] (2026-08-19)` — there is no `0.6.0` section at all.

So the published artefacts, the repo's own version metadata, and the changelog disagree about what version this is. A self-hoster running `0.6.0` has a `package.json` claiming `0.5.0`, and the server health check reports a third number.

## Why

`scripts/release.sh` already does all of this — it is not missing, it is **orphaned**:

> 1. Updates version in root + all workspace package.json files
> 2. Updates the hardcoded version in the server health check
> 3. Updates docker-compose.yml image tags
> 4. Generates a changelog section from conventional commits

The release gate (#505) replaced the manual release flow but **never adopted that script**. Its `release` job tags the graded SHA, writes notes, creates the GitHub Release and dispatches the image build. It does not touch versions or the changelog, and nothing else does either. The prep step simply fell out of the process when the process changed.

## Proposed: a `prepare` stage in the gate

Add a stage between `verify` and `release`:

- Runs the version/changelog prep against the graded SHA
- Commits `chore: release vX.Y.Z` to `main`
- The `release` job then tags **that** commit, not the pre-bump one

### The ordering constraint that makes this non-trivial

The gate deliberately tags the **exact SHA the suite graded** (`needs.suite.outputs.sha`), and re-verifies it before tagging:

```
if [ "$HEAD_SHA" != "$SHA" ]; then
  echo "::error::checked out $HEAD_SHA but the graded commit was $SHA — refusing to tag"
```

That guard exists for a good reason and must not be weakened. A prep commit changes the SHA, so the design has to decide explicitly:

- **tag the prep commit** — then the tagged tree is not the graded tree, and the guard needs to compare "graded SHA is the parent" rather than equality; or
- **tag the graded commit and land the bump separately** — the tag then points at a tree whose `package.json` is stale, which is today's behaviour minus the confusion; or
- **bump before grading** — the suite grades the release commit itself, which is cleanest but means the version bump precedes the decision to release.

The third is the most honest and probably the right answer, but it reorders the workflow. Whichever is chosen, the invariant "we tagged what we tested" has to remain provable.

## Acceptance

- [ ] A cut release leaves every workspace `package.json` on the released version
- [ ] `CHANGELOG.md` gains a section for the released version
- [ ] The server health check reports the released version
- [ ] The relationship between the graded SHA and the tagged SHA is explicit, enforced, and documented
- [ ] `0.6.0` and `0.6.1` are backfilled in `CHANGELOG.md`

santthosh:
Closes #549. Part of #575.

## The bug

**v0.6.0 shipped with every `package.json` still reading `0.5.0`**, and there is no `0.6.0` changelog section. `scripts/release.sh` does all of this — it was simply **orphaned**: #505 replaced the manual flow and never adopted it.

## Option 3 — bump before grading

The gate tags the exact SHA the suite graded and re-verifies before tagging:

```
if [ "$HEAD_SHA" != "$SHA" ]; then
  echo "::error::checked out $HEAD_SHA but the graded commit was $SHA — refusing to tag"
```

A bump commit changes that SHA, so the invariant needs preserving deliberately. Bumping **first** is the only ordering where the tree that was *tested* and the tree that was *tagged* are byte-identical — the guard stays a plain equality check rather than becoming parentage arithmetic, and the tag never points at a tree whose `package.json` is stale.

New flow: `prepare` → `suite` → `verify` → `release`.

## Two things the issue did not anticipate

**1. `release.sh` missed a package.** Its list was hardcoded and `packages/mcp` was never added — which is why it still reads `0.1.0` while everything else reads `0.5.0`. **Wiring the script in unchanged would have shipped a release that bumped 11 of 12 packages and looked correct.** The list is now derived by `find`, so it cannot drift again.

**2. A prep commit to `main` fires the deploy pipeline.** Its E2E gate wants the **same `e2e-fixtures` lock** the release suite holds, with `cancel-in-progress: false`. And `package.json` is *unmapped* in the impact map, so it resolves to the pessimistic `ALL` — a full 48-fixture suite (~$3, ~50 min) queued behind the release's own suite, grading the identical commit.

The deploy gate now **skips a release-prep commit**. That is *not* an unverified deploy — the release gate grades that exact SHA moments later with the same suite, and the comment says so.

Keyed on the **committer**, not the commit message: a message is user-supplied text and anyone could write `chore: release` to skip the gate. There is a test asserting it does not key on the message.

Mapping `package.json` to "no fixtures" would have been the smaller change and is **wrong** — the same file carries dependency changes, which can change anything.

## Tests

11, at the **workflow level**, because that is where the failure lived — the script worked fine, nothing called it:

- `prepare` exists and runs before `suite`
- it calls `release.sh --no-git`, so the gate owns git and there is no second tag path
- **`suite` grades `needs.prepare.outputs.sha`**, not the input ref — the crux of option 3
- the tag guard is still plain equality, so abandoning option 3 fails a test rather than passing silently
- prepare refuses any ref but `main`, and fails if `release.sh` changed nothing
- the deploy gate skips on committer, not message, and still honours both existing switches
- `release.sh` derives its package list

## Also

Backfills the missing **v0.6.0 (57 commits)** and **v0.6.1 (12 commits)** changelog sections, marked as backfilled rather than presented as contemporaneous.

## Accepted cost

If the suite fails after preparing, `main` briefly carries a bump for a release that did not happen. Visible, recoverable, and cheaper than tagging a tree nobody graded. Whether that should auto-roll-back is deliberately out of scope — worth its own issue if it proves annoying.

## Unblocks

**#550** (release notes say what changed) needs the same commit-range generator this now runs, and **#552** should document the finished process.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -311,8 +311,15 @@ jobs:
     # `cancel-in-progress: false`, so this one would queue behind the release's
     # suite and drive the shared fixtures repo afterwards.
     #
-    # Keyed on the committer, not the message: a message is user-supplied text
-    # and anyone could write `chore: release` to skip the gate.
+    # Keyed on the author email rather than the commit message. Both are
+    # settable by anyone who can push to `main`, so this is NOT a security
+    # boundary — and the earlier claim that it was one overstated it. Anyone
+    # with push access can already dispatch this workflow with
+    # `skip_e2e_gate: true`, so nothing is being protected here.
+    #
+    # It is a narrower ACCIDENT filter: a human writing "chore: release v1.2.3"
+    # by hand is plausible, whereas a human committing as
+    # release-gate@mergewatch.ai is not. That is the whole claim.
     if: >-
       vars.E2E_GATE_ENABLED == 'true'
       && github.event.inputs.skip_e2e_gate != 'true'
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -303,7 +303,20 @@ jobs:
     name: E2E gate (dev)
     runs-on: ubuntu-latest
     needs: deploy-dev
-    if: vars.E2E_GATE_ENABLED == 'true' && github.event.inputs.skip_e2e_gate != 'true'
+    # #549 — a release-prep commit is excluded, and this is NOT an unverified
+    # deploy. The Release Gate pushes `chore: release vX.Y.Z` to main and then
+    # grades that exact SHA itself, moments later, with the same suite. Running
+    # here as well would grade the identical commit twice — and worse, the two
+    # runs contend for the `e2e-fixtures` group below with
+    # `cancel-in-progress: false`, so this one would queue behind the release's
+    # suite and drive the shared fixtures repo afterwards.
+    #
+    # Keyed on the committer, not the message: a message is user-supplied text
+    # and anyone could write `chore: release` to skip the gate.
+    if: >-
+      vars.E2E_GATE_ENABLED == 'true'
+      && github.event.inputs.skip_e2e_gate != 'true'
+      && github.event.head_commit.author.email != 'release-gate@mergewatch.ai'
     # This group protects a RESOURCE, not this workflow (#506).
     #
     # The resource is the `mergewatch/fixtures` repo. Every suite run opens real
--- a/.github/workflows/release-gate.yml
+++ b/.github/workflows/release-gate.yml
@@ -61,8 +61,88 @@ permissions:
   actions: write
 
 jobs:
+  # ── prepare ────────────────────────────────────────────────────────────────
+  # #549 — bump versions and the changelog BEFORE grading.
+  #
+  # The gate tags the exact SHA the suite graded and re-verifies that before
+  # tagging. A version-bump commit changes the SHA, so the invariant has to be
+  # preserved deliberately. Bumping first is the only ordering where the tree
+  # that was TESTED and the tree that was TAGGED are byte-identical — the guard
+  # below stays a plain equality check rather than becoming parentage
+  # arithmetic, and the tag never points at a tree whose package.json is stale.
+  #
+  # That is not hypothetical: v0.6.0 shipped with every package.json still
+  # reading 0.5.0, and there is no 0.6.0 changelog section, because #505
+  # replaced the manual flow and never adopted scripts/release.sh.
+  #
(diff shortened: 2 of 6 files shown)


santthosh:
Both findings addressed. The critical was real, and the mechanism was **worse than described**.

## The critical

The review said: the push fails → the job fails → downstream gets an empty ref → checkout falls back to the default branch.

The middle step is wrong, and that makes it worse. The step used `set -uo pipefail` with **no `-e`**, so a failed `git push` did not abort at all — execution continued, `SHA` was captured from a local-only commit, the output was written, and the step exited **0** because the last command succeeded. **The job would have gone green having pushed nothing.**

Fixed at three layers:

1. `set -euo pipefail` — any failure aborts
2. An explicit check that `origin/main` actually points at the pushed SHA, so a push that "succeeded" but landed elsewhere cannot be graded and tagged as if it were on main
3. The reviewer's suggestion, kept as belt-and-braces: `suite` refuses an empty prepared SHA. That one matters on its own — `actions/checkout` treats an empty `ref` as the **default branch**, so a missing output would silently grade the pre-bump tree and the release would tag something the suite never saw

## The warning — and an overclaim of mine it exposed

My comment said keying the deploy-gate skip on the author email rather than the commit message meant "anyone could write `chore: release` to skip the gate" was avoided. **That overstated it.** Both fields are settable by anyone who can push to `main` — and anyone who can push can already dispatch the deploy workflow with `skip_e2e_gate: true`. It is not a security boundary and I should not have implied one.

What it actually is: an **accident filter**. A human writing "chore: release v1.2.3" by hand is plausible; a human committing as `release-gate@mergewatch.ai` is not. The comment and the test now say that instead.

## Tests

13 (up from 11). Two added for the critical: `set -euo pipefail` plus the remote verification, and the suite's empty-SHA guard.

Forced gate green again — build 12/12, typecheck 20/20, test 21/21, `Cached: 0`.

Issue #563, opened 2026-09-05
Re-reviewing the same commit is never billed: the Stripe idempotency key repeats with a different amount


santthosh:
## What happens

Re-reviewing the same commit fails to bill, and says so only in a warning:

```
Failed to debit Stripe customer balance: Keys for idempotent requests can only be used
with the same parameters they were first used with. Try using a key other than
'review-billing-141442688-1907#3219620' if you meant to execute a different request.
```

Observed in production on a re-review triggered by the Re-run button (fixtures#1907), immediately after #558 made that button work.

## Cause

The idempotency key is built from **installation + PR + commit SHA**:

```
review-billing-{installationId}-{prNumber}#{headSha}
```

That is correct for its original purpose — stopping a redelivered webhook from double-charging one review. But it does not distinguish a *redelivery* from a *legitimate second review of the same commit*, and those now both happen:

- **Re-run button** (#558) — re-reviews the same SHA on demand
- **`@mergewatch review`** — same
- **Any retry** after a transient failure

Stripe rejects the second call because the key was first used with a different amount, so the debit silently does not happen. The review is delivered and not paid for.

## Why it matters more now

This is pre-existing, but the Re-run button was inert until #558 shipped today. A path that was effectively unreachable is about to become ordinary, and every use of it is an unbilled review.

## Fix

The key must be unique per *billable event*, not per commit, while still absorbing redeliveries. Options:

1. **Include the review's identity** — the review record's id or `prNumberCommitSha` plus an attempt counter — so a genuine second review gets a genuine second key, and a redelivery of the same review reuses one.
2. **Key on the SQS message id**, which is stable across a redelivery of the same job and distinct between jobs. Closest to the actual semantics of "one charge per job".

Option 2 looks cleaner, but needs checking against the direct-invoke fallback path, which has no message id.

## Do not

Drop idempotency, or make the key random. The original protection is real: a redelivered webhook must not double-charge. This needs a *better* key, not no key.

## Also worth deciding

The failure is a `console.warn` and nothing else — the review proceeds and the customer is not charged. Whether an unbillable review should still be delivered is a product decision, but it should at minimum be visible as an error and countable, rather than a warning nobody reads. Same family as #548.

## Acceptance

- [ ] A second review of the same commit is billed
- [ ] A redelivered webhook for the SAME review still does not double-charge
- [ ] A billing failure is surfaced as an error, not a warning
- [ ] Covered by a test that distinguishes redelivery from re-review

[...]

PR #580, opened 2026-09-08
feat(release): release notes say what changed, not just how it was tested (#550)

### LINKED ISSUE #550: Release notes say how it was tested, never what changed

santthosh:
Closes #550. Part of #575.

## What the notes said

```
Released from `main` (commit `d065b724`) via the release gate (#505).

Graded fixture suite:

47 passed · 0 failed · 0 ungraded · 1 skipped · 0 errored

Manual fixtures verified and approved by @santthosh.
```

That is a **provenance record**, and a good one — it is the evidence this gate exists to produce. But a reader deciding whether to upgrade learns how thoroughly it was graded and **nothing about what it contains**. v0.6.1's notes omitted all eight of its fixes, two of which had made features silently non-functional (#526, #544).

## One generator, two consumers

The changelog logic was inlined in `release.sh`. Extracted to `scripts/changelog-section.sh`, which both `release.sh` (for `CHANGELOG.md`) and the release job (for the notes) now call.

**Two implementations of "what changed in this release" would disagree the first time either was touched** — and the notes are the copy a user reads. There is a test asserting `release.sh` no longer inlines `git log --grep`.

`--no-heading` for the notes, because a GitHub Release's title already *is* the version.

## What changed leads; provenance is kept in full

```
## What changed

### Features
- feat(core): …  (#565) (8aaa02e)

### Bug Fixes
- fix(core): …  (#547) (d065b72)

---

Released from `main` (commit `…`) via the release gate (#505).
Graded fixture suite: …
Manual fixtures verified and approved by @…
```

**Issue links come free.** The repo's commit convention already carries them in the subject — `fix(core): … (#544) (#547)` — so GitHub autolinks them and nothing has to construct a URL.

## The "not verified" criterion was already met

`steps.grade.outputs.summary` greps `passed ·|NOT VERIFIED`, so the count of fixtures *not* run already reached the notes. I checked rather than assuming, and added a test pinning it — a green grade without that count would recreate the coverage illusion the gate exists to remove.

## Empty range

A release with no conventional commits says so plainly:

```
_No feature or fix commits in this range (3 commit(s) since v0.6.1)._
```

An empty section would read like the generator broke. A docs-only or chore-only release is a real case.

## Tests

11 added, including both generator behaviours exercised against the **real repo history** rather than fixtures: grouping with short SHAs, issue numbers present, the empty-range message, the heading emitted for CHANGELOG and suppressed for notes, and `v0.7.0` / `0.7.0` treated identically.

## Limit

Like #549, the notes path only executes during an actual release cut. These tests assert the wiring and the generator's output; **the first real proof is cutting v0.7.0.**

### DIFF
--- a/.github/workflows/release-gate.yml
+++ b/.github/workflows/release-gate.yml
@@ -447,14 +447,22 @@ jobs:
           #
           # Same generator as CHANGELOG.md. --no-heading because a GitHub
           # Release's title already IS the version.
-          CHANGES="$(scripts/changelog-section.sh "$VERSION" --no-heading || true)"
+          # `|| true` would conflate two different things: the generator
+          # FAILING, and there being nothing to report. The generator already
+          # emits an explicit "no feature or fix commits in this range" line
+          # for the second, so an empty result here can only mean the first —
+          # and silently printing a placeholder would ship notes that look
+          # complete. Capture the status and fail the release instead.
+          CHANGES=""
+          gen_rc=0
+          CHANGES="$(scripts/changelog-section.sh "$VERSION" --no-heading)" || gen_rc=$?
+          if [ "$gen_rc" -ne 0 ] || [ -z "$CHANGES" ]; then
+            echo "::error::changelog-section.sh failed (rc=$gen_rc) — refusing to publish notes that omit what changed"
+            exit 1
+          fi
           {
             printf '## What changed\n'
-            if [ -n "$CHANGES" ]; then
-              printf '%s\n' "$CHANGES"
-            else
-              printf '\n_Changelog generation produced nothing — see the commit range._\n'
-            fi
+            printf '%s\n' "$CHANGES"
             printf '\n---\n\n'
             printf 'Released from `%s` (commit `%s`) via the release gate (#505).\n\n' "$REF" "$SHA"
             printf 'Graded fixture suite:\n\n'
--- a/.github/workflows/release-gate.yml
+++ b/.github/workflows/release-gate.yml
@@ -439,7 +439,23 @@ jobs:
           longest=$(printf '%s' "$GRADED" | { grep -oE '`+' || true; } \
             | awk '{ if (length($0) > m) m = length($0) } END { print m+0 }')
           fence=$(printf '%*s' "$(( longest < 3 ? 3 : longest + 1 ))" '' | tr ' ' '`')
+          # #550 — what CHANGED goes first. The provenance block below is the
+          # evidence this gate exists to produce and is kept in full, but a
+          # reader deciding whether to upgrade needs the changes, and the notes
+          # used to omit them entirely: v0.6.1's said nothing about its eight
+          # fixes, two of which had made features silently non-functional.
+          #
+          # Same generator as CHANGELOG.md. --no-heading because a GitHub
+          # Release's title already IS the version.
+          CHANGES="$(scripts/changelog-section.sh "$VERSION" --no-heading || true)"
           {
+            printf '## What changed\n'
+            if [ -n "$CHANGES" ]; then
+              printf '%s\n' "$CHANGES"
+            else
+              printf '\n_Changelog generation produced nothing — see the commit range._\n'
+            fi
+            printf '\n---\n\n'
             printf 'Released from `%s` (commit `%s`) via the release gate (#505).\n\n' "$REF" "$SHA"
             printf 'Graded fixture suite:\n\n'
             printf '%s\n%s\n%s\n\n' "$fence" "$GRADED" "$fence"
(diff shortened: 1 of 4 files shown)


santthosh:
Both fixed. One is a straight hit; on the critical, the finding is worth acting on but its stated mechanism does not hold — recording that rather than quietly applying the patch.

## The critical — right to flag, wrong about why

The finding says `git log $RANGE` lets a value like `HEAD; malicious-command` "break the intended command boundary". **It does not.** An expanded variable is an *argument*; bash does not re-parse it for metacharacters. Verified directly:

```
$ R="HEAD; echo INJECTED"; git log $R --oneline -1
fatal: ambiguous argument 'HEAD;': unknown revision or path not in the working tree.
```

git receives one bad ref and rejects it. There is no shell execution path.

**What is real underneath, and why I fixed it anyway:**

- a `--since` starting with `-` would be read by git as an **option** rather than a revision — that is argument injection proper, and the quoting alone would not stop it
- whitespace or globs would word-split or expand into a malformed invocation

So: `"$RANGE"` quoted, `--` closing the revision list, and `--since` validated against a plain-ref pattern. Rejects `--output=/tmp/pwn`, `a b`, `x;y`; still accepts `v0.6.1` and `HEAD~1`.

Worth stating the exposure honestly: **no caller passes `--since`** — `release.sh` and the workflow both pass only `$VERSION`. This is defence for a flag that exists to be used later, not a fix for a live path.

## The warning — a straight hit, and mine

`CHANGES="$(scripts/changelog-section.sh ... || true)"` conflated two different conditions: the generator **failing**, and there being **nothing to report**.

The generator already emits an explicit `_No feature or fix commits in this range_` line for the second case — so an empty result here can only mean the first. Printing a placeholder would have shipped release notes that look complete while omitting everything the release contained, which is the exact failure #550 exists to fix.

It now captures the status and fails the release:

```
::error::changelog-section.sh failed (rc=$gen_rc) — refusing to publish notes that omit what changed
```

Same family as the swallowed errors in #545 and #563.

## Tests

15 (up from 11). Four added: no `|| true` masking, the range quoted and `--`-terminated, bad refs rejected, ordinary refs still accepted.

Forced gate green — build 12/12, typecheck 20/20, test 21/21, `Cached: 0`.

santthosh:
**Build & Test failed in CI while my local gate passed — and the difference was my tests, not the code.**

They invoked the generator against **this repo's own history** (`--since v0.6.1`). `actions/checkout` clones shallow and **without tags**, so that ref does not exist in CI. My clone has it; the runner's does not.

I had explicitly defended running against real history rather than fixtures, on the grounds that a mocked `git log` cannot catch an invocation that silently returns nothing. That instinct was right — the implementation was wrong, because it made the tests depend on whatever the checkout happens to contain.

They now build a throwaway repo with known commits and a tag, so the real `git log` still runs but nothing depends on the outer clone.

**Verified rather than assumed this time**, since "green locally" is exactly what misled me:

- zero test calls run in the repo root (all use the temp repo)
- the script itself still behaves in a **shallow, tagless clone** — checked by cloning this repo with `--depth 1 --no-tags` and running it there

Forced gate green — test 21/21.

Worth noting the review was 5/5 on the same commit that failed Build & Test. Green checks and a clean review are different signals, and neither is the other.

Issue #576, opened 2026-09-08
Full pre-release audit of README.md and the Mintlify docs site


santthosh:
Part of #575.

**A full audit of README.md and the entire docs site before every release** — every factual claim checked against the codebase, not a linter on selected facts.

The release gate grades fixtures and waits for manual verification of behaviour. **Nothing looks at the documentation at all**: `grep -rn "docs-site" .github/workflows/` returns nothing. 63 `.mdx` pages plus the README have never been built, linked-checked, or read against the code in CI.

## Two drifts found in ten minutes of looking — both invisible to a linter

**1. Permissions — wrong, on the page a security reviewer reads.**

`docs-site/github-app/permissions.mdx:15`:
> \| **Contents** \| `read` \| … **MergeWatch never writes to repository contents.**

The App was changed to `Contents: Read and write` for #545 and the grant accepted. `docs/permissions.md` (#546) says so; this page says the opposite. The screenshot `images/repo.contents-permission.png` shows "Read-only" too.

**2. `maxFindings` — the sentence never changed, but its truth did, twice.**

`docs-site/configuration/mergewatch-yml.mdx:133`:
> \| `maxFindings` \| `number` \| `25` \| Maximum number of findings posted per review.

That was **false** until #569 — `maxFindings` was only substituted into the orchestrator prompt, never enforced. It is now **true but incomplete**: custom-agent findings are exempt (#385), so it is not an absolute ceiling.

This is the one that shows why a fact-checker is not enough. **No link check, nav check, or permissions assertion would flag that sentence.** It reads perfectly. It was simply describing behaviour the product did not have — and now describes behaviour the product has *almost*.

## What a full audit covers

| | |
|---|---|
| **Factual accuracy** | every claim about config, defaults, permissions, pricing, deployment, behaviour, checked against the code that implements it |
| **Completeness** | did this release ship anything the docs do not mention? #558's Re-run button, #563's re-review billing, #569's cap semantics |
| **Staleness** | content describing behaviour that changed or was removed |
| **Internal consistency** | pages contradicting each other, or the README (which contradicts *itself* on Docker publishing — #551) |
| **Executability** | commands and examples actually run; referenced scripts exist |
| **Visual** | screenshots still match the UI they claim to show |
| **Structural** | links resolve, nav resolves both ways, frontmatter valid |

Only the last row is mechanical. The rest needs something that can read prose against code.

## How to make it actually happen

[...]

PR #599, opened 2026-09-09
fix(release): validate the version before prepare mutates main

santthosh:
Unblocks the v0.6.2 cut, and fixes the defect that broke it.

## What happened

The cut was dispatched as `-f version=0.6.2`. `prepare` **accepted** it — `scripts/release.sh` strips a leading `v` — bumped root and all 12 `package.json` files, wrote a CHANGELOG section, committed, and **pushed to main**. `suite` then rejected the *same input* for not matching `v1.2.3`:

```
##[error]version must look like v1.2.3 (got '0.6.2')
```

`main` was left carrying `chore: release 0.6.2` for a release that was never tagged — the inverse of the v0.6.0 defect #549 fixed, reached from the other direction, and triggerable by a typo in a dispatch.

## Two faults

**1. Ordering.** The only version validation lived in `suite`, which runs *after* `prepare` has written and pushed. Validation is now the first thing `prepare` does — before `release.sh`, long before `git push`.

**2. Disagreement.** `prepare` accepted an input `suite` refused. Both now accept the same two spellings (`v0.6.2` and `0.6.2`), with a test asserting the patterns stay identical so they cannot drift apart again.

## Also in here

- **Reverts the orphaned bump** — main is back to `0.5.0` / `mcp 0.1.0` with no `0.6.2` changelog section, so `prepare` has a clean tree to work from. Without this the re-cut fails on `release.sh changed nothing — is 0.6.2 already the current version?`
- **One finding from the docs audit's first CI run.** I changed *"the Postgres service is named `postgres`"* to `db` earlier today and left the rest of the same sentence telling readers to use `postgres` as the hostname. The audit caught the half-fix — the sentence contradicted itself mid-line.

## On the tests

The first version passed for the wrong reason. The run block's comments mention both `release.sh` and `git push` **while explaining the ordering**, so `indexOf` matched prose and the assertion compared comment positions. They strip comment lines and compare executable order now. Confirmed to fail without the fix.

## What the failed run did prove

Worth recording, because two things ran for the first time and both worked:

- **`prepare` (#549) is correct.** It bumped root + all 12 manifests to `0.6.2`, including `packages/mcp` (`0.1.0` → `0.6.2`, the drift #549's derived list was meant to catch), and wrote a proper CHANGELOG section. Its output was right; only its input validation was missing.
- **The docs audit (#576) works in CI.** 64/64 pages, OIDC → Bedrock succeeded in the gate's context, 22 findings refuted and 4 upheld, **$1.78**. Its `status=findings` path is the one that reached the approval checklist, exactly as designed.

`build`, `typecheck`, full suite, and docs structure all pass.

### DIFF
--- a/.github/workflows/release-gate.yml
+++ b/.github/workflows/release-gate.yml
@@ -117,10 +117,14 @@ jobs:
           # `release.sh` strips a leading `v`, so this job accepted an input
           # `suite` would refuse. Validate first, and accept both spellings so
           # the two jobs cannot disagree again.
-          case "$VERSION" in
-            v[0-9]*.[0-9]*.[0-9]*|[0-9]*.[0-9]*.[0-9]*) ;;
-            *) echo "::error::version must look like v1.2.3 or 1.2.3 (got '$VERSION')"; exit 1 ;;
-          esac
+          # A `case` glob is too loose to gate a mutation: `*` spans anything,
+          # so v1.2.3.4 and v1.2.3junk both matched and would have produced a
+          # bad tag. Anchored regex — optional `v`, exactly three numeric
+          # components, nothing else.
+          if [[ ! "$VERSION" =~ ^v?[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+            echo "::error::version must look like v1.2.3 or 1.2.3 (got '$VERSION')"
+            exit 1
+          fi
 
           # Only `main` can be prepared: the bump commit is pushed there, and
           # pushing it while grading a different ref would tag a tree that was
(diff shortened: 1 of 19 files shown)


santthosh:
Both warnings addressed — and they were right in a way that mattered more here than usual.

A `case` glob is too loose to gate a mutation. `*` spans anything, so `v1.2.3.4` and `v1.2.3junk` both matched:

```
  v1.2.3.4       ACCEPT      <- would have produced a bad tag
  v1.2.3junk     ACCEPT
```

That was tolerable while the check lived only in `suite`, where a bad match just failed the run. This PR moves it in front of `prepare`, which **commits and pushes** — so anything the check waves through is written to main before another job sees it. A loose validation guarding a mutation is a worse thing than a loose validation guarding a report.

Anchored regex at both sites now — optional `v`, exactly three numeric components, nothing else:

```
  v1.2.3  0.6.2  v0.6.2  10.20.30   ACCEPT
  v1.2.3.4  v1.2.3junk  v1..2  1.2  vX.Y.Z  <empty>  v1.2.3-rc1   reject
```

Pre-release suffixes are rejected too. The gate has never cut one, and starting to accept them by accident is not the way to begin.

Added a test enumerating both the accepted and rejected cases, and changed the "same spellings" test to compare the two jobs' validation lines **to each other** rather than to a hand-escaped literal — the property worth pinning is that they are identical, whatever they say.

santthosh:
Fixed. `steps.find(...)` returns `undefined` when nothing matches, so renaming or removing the `commit` id would throw a `TypeError` — which reads as a broken test rather than as *"the step these assertions guard is gone"*.

The helper now asserts the step resolved, with a message naming the missing id.

Worth admitting this is the **same defect I fixed in `docs-structure.test.ts` earlier today** — reading a workflow node without first checking it resolved. Same shape, same file-family, a few hours apart.

It matters more here than it did there: these assertions are the only thing standing between a typo'd version and a push to `main`. A test that fails confusingly is one someone eventually deletes, and deleting this one restores the exact defect this PR exists to fix.

The two remaining semver warnings are carry-forwards the review itself marks as pre-existing — they refer to the `case` glob that this PR replaced with an anchored regex.

Issue #455, opened 2026-08-23
Review skip: .github/** YAML is treated as trivial — CI and deploy workflows are never reviewed


santthosh:
`SKIP_PATTERNS` in `packages/core/src/skip-logic.ts` classifies YAML/CI config as trivial:

```ts
  // Config-only files (version bumps, CI tweaks)
  '**/.github/**',
```

A PR touching **only** files under `.github/` is smart-skipped and never reviewed.

## Evidence from this repo, today

Every workflow change shipped in this session was skipped:

| PR | Files | MergeWatch |
|---|---|---|
| #446 | `.github/workflows/deploy.yml` | `skipping` |
| #448 | `.github/workflows/deploy.yml` | `skipping` |
| #450 | `.github/workflows/deploy.yml` | `skipping` |
| #452 | `.github/workflows/deploy.yml` | `skipping` |
| #454 | `.github/workflows/deploy.yml` | `skipping` |

What was in those diffs:

- **#448** added a job that **blocks production deploys**, and wired in a cross-repo write token (`FIXTURES_TOKEN`).
- **#452** fixed a **command-injection hole** — a `workflow_dispatch` input interpolated via `${{ }}` directly into a `run:` block, in a workflow holding AWS OIDC credentials and that token. Caught by hand; the reviewer never saw the PR.
- **#446** changed concurrency semantics for every deploy.

The file that deploys to production is the one file the reviewer is configured to ignore.

## Why `.yml` is the wrong thing to call trivial

CI and deployment YAML is not documentation — it is **executable configuration with credentials in scope.** GitHub Actions workflows are a well-known home for:

- `${{ }}` template injection into `run:` blocks
- secret exfiltration via a malicious step or a bumped third-party action
- `pull_request_target` misuse
- permission widening (`permissions: write-all`)
- concurrency and gating changes that silently disable protections

None of that is a "version bump or CI tweak". The comment in the source describes the *intent* (skip trivial dependency bumps) but the pattern is far broader than the intent.

The same argument covers `docker-compose.yml`, k8s manifests, `.gitlab-ci.yml`, and `serverless.yml`. None are currently matched by name — but `.github/**` catches the highest-risk instance of the class.

## What stays skippable

This is a narrowing, not a removal. Genuinely ignorable content keeps its exemption:

- **Documentation** — `*.md`, `*.mdx`, `*.txt`, `*.rst`, `docs/**`, `CHANGELOG*`, `LICENSE*`
[...]

Issue #602, opened 2026-09-09
/mergewatch reject on an inline thread produces no confirmation and no disposition record


santthosh:
Found performing manual verification of `40-mergewatch-reject` for the v0.6.2 release cut. **The fixture fails.**

## What was done

1. Applied `40-mergewatch-reject` → [mergewatch/fixtures#2819](https://github.com/mergewatch/fixtures/pull/2819). Both apps reviewed and posted inline threads.
2. Replied on the prod app's inline thread (comment `3969677152`) with the exact command the fixture's README specifies:

```
/mergewatch reject style-disagreement we use snake_case for python here
```

3. Waited ~105s and checked both the thread and the disposition store.

## Expected vs observed

| README expectation | Observed |
|---|---|
| Bot posts a structured confirmation reply | **absent** |
| `FindingDispositionRecord` with `disputeCount = 1` | **absent** — zero rows mentioning PR 2819 in `mergewatch-finding-dispositions-prod` |
| `rejectReasons[0] = { category: 'style-disagreement', text: '…', at: … }` | **absent** |
| Thread NOT auto-resolved | (holds — nothing happened at all) |

## What has been ruled out

- **Not an unsubscribed event.** `GET /app` confirms the prod App subscribes to `pull_request_review_comment`. (Unlike `check_suite` — see #601.)
- **Not undelivered.** `GET /app/hook/deliveries` shows 11 recent `pull_request_review_comment` deliveries, all **200**.
- **Not a permissions failure.** The reply posted fine and the App has `pull_requests: read & write`.

## What is not yet isolated

The prod webhook log for the relevant window shows only `Ignoring unhandled event type: pull_request_review` — no line from `handleReviewCommentEvent`. So either the event for *this specific reply* did not arrive, or it arrived and `shouldHandleReviewCommentEvent` returned false.

That predicate has four gates:

```ts
if (event.action !== 'created') return false;
if (isBotActor(event.sender) || isBotActor(event.comment.user)) return false;
if (event.comment.in_reply_to_id == null) return false;
if (!event.installation?.id) return false;
```

All four look satisfiable by a human reply via `POST /pulls/{pr}/comments/{id}/replies`. Worth checking whether that endpoint sets `in_reply_to_id` on the delivered payload the way a UI reply does — the reply also generated `pull_request_review` events, which suggests GitHub may model an API reply differently from a UI one.

[...]

Issue #601, opened 2026-09-09
The prod GitHub App is not subscribed to check_suite — #558's fix is inert in production


santthosh:
Found while performing the manual verification for the v0.6.2 release cut.

## The App's actual event subscriptions

Queried directly from GitHub as the App (`GET /app` with a JWT minted from the prod private key):

```
subscribed: check_run, issue_comment, pull_request,
            pull_request_review, pull_request_review_comment,
            pull_request_review_thread

MISSING:    check_suite
            installation
```

## Why `check_suite` matters

#558 fixed a production defect where GitHub's **Re-run all checks** button did nothing, because that button fires `check_suite.rerequested`, not `check_run.rerequested`. The fix added the handler and the dispatch case, and it is correct — I read it.

**But the App was never subscribed to `check_suite`.** GitHub does not deliver events an App has not subscribed to, so the handler cannot run. The button is still dead in production, for every user, on every PR — the exact wording #558 used to describe the original bug.

This is the second time this feature has shipped broken, and the second failure is worse than the first: the first was a wrong event name in code, which a reader could catch. This one is a setting outside the repository that no test, no review, and no deploy can see.

`E2E-12` (`12-rerun-check`) is the fixture that covers this. It is `MANUAL_ONLY`, and per #515 has never been verified — so nothing has ever exercised the path.

## Why `installation` matters

`handleInstallationEvent` upserts installation and repository records on install, and logs on uninstall. Not subscribed, so it never runs. New installs are recorded only as a side effect of their first PR review, and uninstalls are never observed at all.

## `marketplace_purchase` is a separate story

Also absent from this list, but that is expected and not a defect: Marketplace listing events are configured on the **listing's own webhook**, not the App's event subscriptions. That webhook was set up separately and its ping was confirmed reaching prod (#596).

## The pattern

Three defects this week share a shape — #545 (a permission), #558 (an event name), and now this (an event subscription). All are configuration outside the repo, all are invisible to CI, and all fail by silence rather than error. A green deploy proves the code is right and says nothing about whether it can be reached.

**A test asserting that the App's live subscriptions match what `webhook.ts` dispatches on would have caught all three.** That comparison is mechanical: the dispatch's `case` labels are in the repo, and `GET /app` returns the subscriptions.

## Fix

1. Subscribe the prod App to `check_suite` and `installation`. Same for the dev App.
2. Verify `E2E-12` end-to-end afterwards, since it has never been run.
[...]

PR #603, opened 2026-09-09
feat(core): let a repo declare which of its files are not worth reviewing

davidsun-mindbody:
> **As a maintainer of a repository with non-code files MergeWatch can't know about,**
> **I want** to declare which paths are trivial,
> **so that** editing tooling config or agent skills doesn't cost a full review.

Opening upstream at Santthosh's suggestion — this has been running through review on a downstream fork.

## The gap

Every CI system lets a repository say which paths don't warrant a given job — don't run the unit suite for a README edit, don't build an image for a docs change. That judgement is always repository-specific, because only the repository knows which of its files can affect the thing being built.

MergeWatch already accepts that reviews shouldn't run on everything — that is what `SKIP_PATTERNS` is for. But the list is hardcoded, so it can only encode a global guess: lock files, docs, editor config. It cannot know that in a given repository the review tooling's own config file, an agent-skills directory, or a generated inventory are equally not code. PRs touching only those get a full review and a bill for it.

## What already exists, and why none of it covers this

| Existing lever | What it does | Why it doesn't fit |
|---|---|---|
| `SKIP_PATTERNS` | The built-in trivial list | Hardcoded — a repo can't add to it |
| `includePatterns` | Overrides `SKIP_PATTERNS` | Opts files back **in**; the opposite direction |
| `excludePatterns` | Drops files from the diff sent to agents | The PR is still reviewed, and still billed |
| `rules.ignorePatterns` | Deprecated; folds into `excludePatterns` | Same layer, not a skip |
| `maxFileDiffKB` (#423) | Drops one oversized file's diff section | Diff-filter, and size-based rather than path-based |
| `rules.ignoreLabels` (`skip-review`) | Skips a labelled PR | Manual and per-PR — you must remember, every time |
| `rules.maxFiles` | Skips very large PRs | Size, not path |
| `rules.skipDrafts` | Skips drafts | Unrelated axis |

So the skip decision is reachable by size, by label, by draft status, and by a hardcoded pattern list — but not by a path list the repository controls. #423's reasoning is the closest neighbour and points the same way: *"a pattern list only catches artifacts we have already met."* That argument applies to the skip layer too, and this is the other half of it.

I also confirmed `shouldSkipPR` receives the raw PR file list — `excludePatterns` is not applied beforehand — so there is no indirect route where excluding every file in a PR causes it to be skipped.

## 1. `skipPatterns`

A user list that adds to the built-in one, mirroring `includePatterns` in shape, parsing and precedence:

```yaml
skipPatterns:
  - ".mergewatch.yml"
  - "skills/**"
```

- Adds to `SKIP_PATTERNS` rather than replacing it.
- `includePatterns` still wins, so a file can be forced back into review after being skipped.
- Operates at the PR-skip layer, not the diff-filter layer — it decides whether there is a review, where `excludePatterns` decides what reaches the agents once there is one.

Matched with **`dot: true` for the user list only**. Without it a configured `**/*.yml` silently fails to match `.github/workflows/ci.yml`, because minimatch will not let a wildcard cross a dot-directory by default — exactly the class of file someone would reach for this setting to skip. The built-in list keeps its current matching, so no existing behaviour moves.

## 2. `**/.github/PULL_REQUEST_TEMPLATE*` never matched `.github/pull_request_template.md`

minimatch is case-sensitive, and the lowercase form is what `gh` scaffolds and what most repositories actually commit. The effect is that editing a PR template — prose and forms, not automation, and explicitly intended to be trivial by the pattern above it (#455) — draws a full review.

I added the lowercase entry rather than making the whole list case-insensitive, which would have changed matching for every other pattern too.

## Where this came from

Wiring MergeWatch into a repository whose CI already declares which paths cannot affect the build — docs, agent skills, tooling config. Those two lists answer the same question, and there was no way to reuse the answer. The template-casing bug surfaced from the same work: a one-line edit to a PR template kept pulling reviews nobody wanted.

## Testing

Tests added in `skip-logic.test.ts` cover: a configured pattern skipping a PR; the same PR being reviewed when unconfigured; the dot-directory case; a mixed skip-pattern-plus-real-code PR still being reviewed; `includePatterns` winning over `skipPatterns`; the lowercase template; and `extractSkipPatterns` rejecting malformed input.

These ran green on a downstream fork's CI, including `Build & Test`, on the same commit content.

## Files

- `packages/core/src/skip-logic.ts` — `extractSkipPatterns`, third parameter on `shouldSkipPR`, lowercase template pattern
- `packages/core/src/config/defaults.ts` — field + default
- `packages/core/src/github/client.ts` — YAML parsing
- `packages/core/src/index.ts` — export
- `packages/server/src/review-processor.ts`, `packages/lambda/src/handlers/review-agent.ts` — both transports pass it through
- `README.md` — documents `skipPatterns` next to `excludePatterns`, and labels what `excludePatterns` actually does, since the distinction is easy to miss

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,9 +164,17 @@ rules:
   reviewOnMention: true
   skipDrafts: true
 
+# Files whose contents are dropped from the diff sent to the agents.
 excludePatterns:
   - "**/*.lock"
   - "**/dist/**"
+
+# Files that count as trivial, so a PR touching only these is skipped without
+# a review at all. Adds to the built-in list; `includePatterns` overrides both.
+# Use it for the non-code files only your repo knows about.
+skipPatterns:
+  - ".mergewatch.yml"
+  - "skills/**"
 
 # Review tone: collaborative | direct | advisory
 tone: collaborative
--- a/packages/core/src/config/defaults.ts
+++ b/packages/core/src/config/defaults.ts
@@ -175,6 +175,12 @@ export interface MergeWatchConfig {
    * PR gets reviewed at all when every file is otherwise trivial.
    */
   includePatterns: string[];
+  /**
+   * Patterns that ADD to the built-in trivial list. Decides whether a PR is
+   * reviewed at all; `excludePatterns` decides what reaches the agents once it
+   * is. `includePatterns` overrides both.
+   */
+  skipPatterns: string[];
   /** Minimum severity to report: 'info' | 'warning' | 'critical' */
   minSeverity: 'info' | 'warning' | 'critical';
   /**
@@ -259,6 +265,7 @@ export const DEFAULT_CONFIG: MergeWatchConfig = {
     '**/*.wasm',
   ],
   includePatterns: [],
+  skipPatterns: [],
   minSeverity: 'info',
   minConfidence: 75,
   maxFindings: 25,
--- a/packages/core/src/github/client.ts
+++ b/packages/core/src/github/client.ts
@@ -1411,6 +1411,9 @@ export function parseRepoConfigYaml(content: string): Partial<MergeWatchConfig> 
     if (Array.isArray(parsed.includePatterns)) {
       config.includePatterns = parsed.includePatterns.filter((p: unknown) => typeof p === 'string');
     }
+    if (Array.isArray(parsed.skipPatterns)) {
+      config.skipPatterns = parsed.skipPatterns.filter((p: unknown) => typeof p === 'string');
+    }
     if (parsed.agents && typeof parsed.agents === 'object') {
       const a = parsed.agents as Record<string, unknown>;
       config.agents = {
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -254,7 +254,7 @@ export { invokeWithFileFetching, FILE_REQUEST_INSTRUCTION } from './context/agen
 export type { FileFetchOptions, AgenticInvokeResult } from './context/agentic-fetcher.js';
 
 // ─── Skip logic ─────────────────────────────────────────────────────────────
-export { shouldSkipPR, shouldSkipByRules, isAutoReviewOff, extractIncludePatterns, SKIP_PATTERNS } from './skip-logic.js';
+export { shouldSkipPR, shouldSkipByRules, isAutoReviewOff, extractIncludePatterns, extractSkipPatterns, SKIP_PATTERNS } from './skip-logic.js';
 export type { RulesSkipKind, RulesSkipResult } from './skip-logic.js';
 
 // ─── Agent-authored PR detection ────────────────────────────────────────────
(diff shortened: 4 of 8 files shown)


santthosh (approved):
Approving. I verified this rather than reading it — checked out `f0c96b1` and ran it locally.

| Check | Result |
|---|---|
| `pnpm run build` / `typecheck` | ✅ |
| Full suite with `TURBO_FORCE=true` (`Cached: 0`) | ✅ 1,276 core / 253 lambda |
| `config-key-usage.test.ts` (#310 tripwire) | ✅ |
| Behaviour, against the built code | ✅ all four cases |

```
lowercase-template-only PR -> SKIPPED (new)
mixed w/ real code         -> REVIEWED
configured skipPattern     -> SKIPPED (new)
includePatterns wins       -> REVIEWED
```

## What stands out

**Both transports are wired.** `review-agent.ts` and `review-processor.ts` each call `extractSkipPatterns` and pass it through. That is the exact failure mode this repo has hit four times in the last week — a stage variable not threaded (#544), a wrong event name (#558), an unsubscribed event (#601), a handler never reached (#602). Every one shipped looking correct and did nothing. Getting it right in both runtimes, without that history, is the single most valuable thing in this diff.

The layering argument is also right, and it is the part that is easy to get wrong: `excludePatterns` decides what reaches the agents once a review is happening; this decides whether there is a review. Those are genuinely different questions and the existing config could only express the first.

I compared `extractSkipPatterns` against `extractIncludePatterns` line by line — it mirrors the shape, parsing and precedence as claimed. The default parameter means no existing call site moves.

Restraint worth naming: fixing the one lowercase pattern rather than making the whole list case-insensitive. The broad fix would have silently changed matching for every other pattern, and nothing would have caught it.

## Two notes, neither blocking

**1. This is a behaviour change, not only a bug fix.** A PR touching only `.github/pull_request_template.md` is now skipped where it was previously reviewed — confirmed above. It is the right change; the uppercase pattern above it plainly intended to cover the lowercase form. But it should appear in the release notes as a behaviour change, because a repo that relied on those reviews will notice.

**2. `config.skipPatterns` is written and never read.** `client.ts` parses it into the merged config, but both runtimes consume `extractSkipPatterns(yamlConfig)` off the raw YAML instead. That is the #310 shape — except `includePatterns` does exactly the same thing, so it is pre-existing convention that this PR mirrors correctly rather than a defect it introduces. Filed separately as #611; not this PR's to carry.

## A correction on my own review

My first run of the config-key tripwire reported a **failure** on `skipPatterns`. That was my mistake — the test derives `REPO_ROOT` from `process.cwd()` and must run from `packages/core`; I ran it from the repo root. Run correctly it passes. I mention it because I nearly filed a false defect against your PR.

Relatedly, my first `pnpm run test` reported PASS with 9 of 21 tasks **cached**. Only the `TURBO_FORCE=true` run is evidence.

## Where this lands

Directly on **#519** (cost scales with commit churn), a P1 customer-feedback item in v0.6.3 — this gives a repository a lever against exactly that, which the hardcoded list could not.

Thanks for the write-up. The table of existing levers and why each one does not fit saved me the work of checking, and it was accurate when I did.

Issue #600, opened 2026-09-09
Dropping a cluster representative silently discards every finding merged into it


santthosh:
Diagnosed from the repaired decision ledger (#594, fixed by #595 + #598). This is the product defect those ledger fixes made visible.

## What happens

Pipeline order is `dedupeCrossAgentByLine` (FP-C, `reviewer.ts:3161`) → `runOrchestratorAgent` (`:3194`). FP-C merges same-line cross-agent findings into a representative; the orchestrator then judges the merged set.

**When the orchestrator drops a representative, everything merged into it is discarded with it.** The absorbed siblings are not reinstated, not reconsidered, and not reported.

## Observed, twice, on `mergewatch/fixtures#2699`

Same commit, same `CodeSha256`, same model, two installations:

| | trace | rendered |
|---|---|---|
| prod | 2 dropped, 6 merged, **1 surfaced** | 🔴 **1/5** — both criticals |
| dev | 2 dropped, 5 merged, **0 surfaced** | 🟢 **5/5 — "No issues found in the diff"** |

In dev, FP-C merged **five** findings into the representative *"Missing authorization check for admin endpoint"*. The orchestrator dropped that representative. All six disappeared — including a missing-authorization critical raised at **confidence 95 by three independent agents** (`security`, `bug`, `test-coverage`).

In prod the orchestrator kept a representative, it took the `— and N related` rename, and it surfaced. One LLM call's judgement decided whether six findings existed.

## Why this is the wrong shape

**A finding is safest when no other agent agrees with it.** Convergence is what puts a finding into a cluster, and being in a cluster is what makes it vanish. The strength of the signal is what destroys it.

That inverts what FP-C's merge is presented as doing. Merging is deduplication — collapsing several descriptions of one defect into the clearest one. It is not supposed to make those findings share a fate, and nothing downstream knows that dropping one item now drops six.

The user-visible result is the worst available: not a missed finding, but a confident **"No issues found in the diff"** on a diff containing an unauthenticated admin endpoint.

## Fix — needs a decision, both options have blast radius

1. **Reinstate on drop.** If the orchestrator drops a representative, return its absorbed siblings to the pool as independent findings. Faithful to intent, but can resurrect a cluster the orchestrator meant to reject wholesale.
2. **Refuse the drop.** Do not let the orchestrator drop a representative that absorbed anything; require it to drop the members individually. Simpler, but constrains a stage that exists to make exactly this kind of judgement.

[...]

PR #612, opened 2026-09-10
fix(release): make dry_run actually dry — prepare must not push

### LINKED ISSUE #592: dry_run still bumps the version and pushes to main, so it is not a dry run
Found while pre-flighting the v0.6.2 cut — specifically, while checking whether a dry run could safely exercise the new documentation audit (#576) before a real release.

## The gap

`dry_run` guards `verify` and `release`:

```yaml
verify:
  if: ${{ !inputs.dry_run }}
release:
  if: ${{ !inputs.dry_run }}
```

`prepare` has **no such guard**, and it ends with:

```bash
git add -A
git commit -m "chore: release $VERSION"
git push origin main
```

So a dry run bumps every `package.json`, writes a CHANGELOG section, commits, and **pushes to `main`** — then skips the parts that would have turned that into a release. `main` is left claiming a version that was never tagged and never shipped.

That is the inverse of the v0.6.0 defect #549 was filed to fix. #549 solved "tagged without bumping"; this creates "bumped without tagging".

## Why it was not noticed

`prepare` is new in #549, and its own header documents an accepted cost:

> if the suite then fails, `main` briefly carries a bump for a release that did not happen. Visible, recoverable, and cheaper than tagging a tree nobody graded.

That reasoning is sound for a **real** run whose suite fails — an exceptional outcome, and recoverable. It does not transfer to `dry_run`, where skipping the release is the *whole point*. A dry run does not leave a stray bump occasionally; it leaves one **every single time**.

The guard was simply added to the two jobs that existed when `dry_run` was written. `prepare` arrived later and did not inherit it.

## Consequence today

`dry_run` cannot be used. It is the natural way to exercise the gate — including the first CI run of the #576 audit — without cutting a tag, and using it costs a manual revert of `main` afterwards.

## Suggested fix

Not simply `if: ${{ !inputs.dry_run }}` on `prepare` — `suite` and `audit` need its `sha` output. On a dry run it should skip the bump and the push and report the current `main` SHA, so the suite grades the candidate as it stands:

```bash
if [ "${DRY_RUN}" = "true" ]; then
  SHA="$(git rev-parse HEAD)"
  echo "sha=$SHA" >> "$GITHUB_OUTPUT"
  echo "dry run — no bump, no push; grading main as-is at $SHA"
  exit 0
fi
```

Worth stating in the summary too, so a dry run's output cannot be mistaken for a prepared release.

## Acceptance

- [ ] A dry run leaves `main` byte-identical — no bump, no CHANGELOG edit, no push
- [ ] `suite` and `audit` still receive a valid SHA on a dry run and grade the real candidate
- [ ] The step summary distinguishes "prepared for release" from "dry run, nothing prepared"
- [ ] A test asserts `prepare` cannot push when `dry_run` is true — the same shape as the existing release-gate invariant tests

santthosh:
Closes #592. PR 1 of 2; #601's drift test follows.

## The defect

`dry_run` guarded `verify` and `release`. **`prepare` never received the input at all** — so a dry run bumped root and all 12 `package.json` files, wrote a CHANGELOG section, committed, and **pushed to `main`**, then skipped the jobs that would have made it a release.

`main` was left claiming a version that was never tagged — the inverse of the v0.6.0 defect #549 fixed, reachable from the one flag whose entire purpose is to change nothing. **The cautious path was the destructive one.**

## The fix

Bail out before `release.sh` touches a file, and long before the push:

```bash
if [ "$DRY_RUN" = "true" ]; then
  SHA="$(git rev-parse HEAD)"
  echo "sha=$SHA" >> "$GITHUB_OUTPUT"
  …
  exit 0
fi
```

**Still emitting `sha` matters.** Exiting without it would leave `suite` grading an empty string — the gate would go green having checked nothing, which is a worse bug than the one being fixed. `suite` and `audit` both still `needs: prepare`, so they grade `main` as it stands.

The two summaries now say which happened — **"Prepared for release"** vs **"Dry run — nothing prepared"**. A dry run's output could otherwise be mistaken for a cut, which is the same confusion relocated.

## On the tests, and a mistake worth naming

Seven tests. Initially only three failed without the fix — and **two of the ordering tests passed on the broken code**:

```ts
expect(run.indexOf('DRY_RUN')).toBeLessThan(run.indexOf('git push'));
```

With the guard absent, `indexOf` returns `-1`, and `-1 < anything` holds. They asserted the ordering on code that had **no guard at all**. They assert presence first now:

```ts
expect(at, 'the prepare step never mentions DRY_RUN').toBeGreaterThan(-1);
```

**Five of seven now fail without the fix** — verified by reverting it and re-running, not assumed. The remaining two are back-stops (`verify`/`release` still guarded; `suite`/`audit` still depend on `prepare`) which should hold either way.

## What this does not do

It does not make `dry_run` exercise the tag or release path — by definition. What it now gives you is a **safe rehearsal of `prepare`, `suite` and `audit`**, which is what you actually want before a real cut, and what was unavailable when the v0.6.2 cut needed it.

## Verified

`build`, `typecheck`, and the full suite with `TURBO_FORCE=true` (**`Cached: 0`**) all pass.

### DIFF
--- a/.github/workflows/release-gate.yml
+++ b/.github/workflows/release-gate.yml
@@ -101,6 +101,7 @@ jobs:
         env:
           VERSION: ${{ inputs.version }}
           REF: ${{ inputs.candidate_ref }}
+          DRY_RUN: ${{ inputs.dry_run }}
         run: |
           # -e matters here. Without it a failed `git push` does NOT abort:
           # execution continues, SHA is captured from a local-only commit, the
@@ -132,6 +133,33 @@ jobs:
           if [ "$REF" != "main" ]; then
             echo "::error::candidate_ref must be 'main' to prepare a release (got '$REF')"
             exit 1
+          fi
+
+          # #592 — a dry run must be DRY. `dry_run` guarded `verify` and
+          # `release`, but not this job, so a dry run bumped every
+          # package.json, wrote a CHANGELOG section, committed and PUSHED to
+          # main — then skipped the jobs that would have turned that into a
+          # release. main was left claiming a version that was never tagged:
+          # the inverse of the v0.6.0 defect #549 fixed.
+          #
+          # Bail out BEFORE release.sh touches anything, and hand downstream
+          # the SHA main already has, so `suite` and `audit` still grade the
+          # real candidate rather than being skipped.
+          if [ "$DRY_RUN" = "true" ]; then
+            SHA="$(git rev-parse HEAD)"
+            echo "sha=$SHA" >> "$GITHUB_OUTPUT"
+            echo "dry run — nothing prepared; grading main as-is at $SHA"
+            {
+              echo "## Dry run — nothing prepared"
+              echo ""
+              echo "- Requested version: \`$VERSION\` (**not** applied)"
+              echo "- Grading: \`$SHA\` — main as it stands"
+              echo ""
+              echo "No version bump, no CHANGELOG entry, no commit, no push. The suite and"
+              echo "the documentation audit run against the candidate; \`verify\` and"
+              echo "\`release\` are skipped, so no tag is cut."
+            } >> "$GITHUB_STEP_SUMMARY"
+            exit 0
           fi
 
           git config user.name  "mergewatch-release-gate"
@@ -167,7 +195,7 @@ jobs:
           echo "sha=$SHA" >> "$GITHUB_OUTPUT"
           echo "prepared $VERSION at $SHA"
           {
-            echo "## Prepared"
+            echo "## Prepared for release"
             echo ""
             echo "- Version: \`$VERSION\`"
             echo "- Commit: \`$SHA\`"
--- a/packages/lambda/src/release-gate.test.ts
+++ b/packages/lambda/src/release-gate.test.ts
@@ -365,11 +365,26 @@ describe('#592 — a dry run must be dry', () => {
   });
 
   it('summaries distinguish a dry run from a real preparation', () => {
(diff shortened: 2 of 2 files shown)


santthosh:
Both addressed.

**Test placement — real, fixed.** You are right that asserting the two strings merely *exist* was weak: it would have passed if they were swapped, or if both sat in the same branch, which is precisely the confusion the strings exist to prevent. The test now checks position — dry-run summary inside the guard before `exit 0`, prepared summary after it, on the path a dry run never reaches.

I verified it by swapping the two strings in the workflow and confirming the test fails, rather than assuming the new assertion bites.

**Shell injection — mitigated, and I checked the ordering rather than taking the mitigation on trust.** `$VERSION` reaches the summary only after the anchored validation at the top of the same run block:

```bash
if [[ ! "$VERSION" =~ ^v?[0-9]+\.[0-9]+\.[0-9]+$ ]]; then … exit 1; fi
```

That permits an optional `v` and three numeric groups and nothing else — no metacharacters, no whitespace, no substitution. The validation is the first thing in the block, before the ref check and long before the summary, so there is no path where an unvalidated `$VERSION` is interpolated. #599 tightened this from a `case` glob for exactly this reason: the glob accepted `v1.2.3junk`.

Worth noting for the record: this finding **did not appear in the rendered list** — the comment says *"1 finding was dropped or clustered by post-orchestrator filtering"*. It only survived in the summary line. That is #600 in our own review of our own PR: a finding merged into a cluster whose representative was dropped. Useful confirmation that #600 is not fixture-only.

